PR #85, opened 2017-10-30
Deploy new icons to prod

andrewkrug:
* Right size a couple icons.


andrewkrug:
@comzeradd please review on staging and add your +r if this looks OK

PR #124, opened 2018-02-23
Updated URL and authorized groups for community analytics

viorelaioia:


viorelaioia:
@hmitsch , I updated the authorized groups because I noticed that non-nda users are not allowed to access community analytics. They can see the app in sso dashboard, but when trying to access it, they get a forbidden error page. Please let me know if my assumption is not correct and I will update the PR.

andrewkrug:
@viorelaioia can you redo this PR by forking again from master?  We ran into some issues with unicode characters in the apps.yml file. 

I could rebase it for you.  But I want to make sure the new test pipeline is working.

PR #125, opened 2018-02-26
restrict slack to invited community only

andrewkrug:
https://bugzilla.mozilla.org/show_bug.cgi?id=1441201


andrewkrug:
@jdow @gdestuynder I've added you as reviewers on this simple change only because of my discovery of the unicode characters.  

TravisCI will now reject builds of this file that would break authentication or the dashboard inline.  *Do not override merges if travis fails* 

Also note the introduction of a CHANGELOG for this in "keep a changelog format" with semantic versions.  Please bump the version for each change to the config and add a changelog entry going forward.

PR #129, opened 2018-03-01
Take community analytics changes live and more.

andrewkrug:


hmitsch:
Do I still need to approve? If yes, I don't see a button for it. And *yes*, I would approve.

PR #133, opened 2018-03-15
Rename "Legacy Air Mozilla" to "Air Mozilla (Legacy)"

april:
This will put the new and old Air Mozilla icons next to each other on the dashboard. It's not necessarily obvious at first glance that the Legacy air.mo is available anymore.

Thanks!


hidde:
+r

PR #136, opened 2018-03-22
adding syntrio

jdow:


gene1wood:
This is missing the `client_id`. A workaround has been added in #503 but this should be fixed

PR #137, opened 2018-04-06
updating servicenow url

jdow:
This resolves a bug with SN's redirect logic per https://bugzilla.mozilla.org/show_bug.cgi?id=1451865

### DIFF
--- a/apps.yml
+++ b/apps.yml
@@ -705,7 +705,7 @@ apps:
       name: "The Hub"
       client_id: "5gtZrLu2eyAapp1BgQsF11rhdPNt2lGP"
       op: auth0
-      url: "https://auth.mozilla.auth0.com/samlp/5gtZrLu2eyAapp1BgQsF11rhdPNt2lGP"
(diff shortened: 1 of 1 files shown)


gdestuynder:
/me wonders if theres a better way

jdow:
I'm not sure what you mean by a better way. There are a few RPs that specifically don't support IDP-initiated SAML and this appears to be one of them. While it has technically worked in most cases in the past, it appears that SP-initiated SAML is the proper fix for this, and in those cases, even Okta recommends using the RP's SP-initiated SAML URL for their dashboard. I believe Expensify is set up the same way.

gdestuynder:
yeah, its just a little odd for the user if the RP doesnt redirect them (=wondering if theres a better way)

jdow:
It seems that the RP does redirect them.

gdestuynder:
yes, in this case, but theres no garantee. some providers will just show you a prompt that says "click here to login with your company's SSO" or worse :)
anyway, i dont have any other way to do this

PR #142, opened 2018-04-20
update expensify logo and update airmo group config

jdow:


gdestuynder:
group change request  was https://bugzilla.mozilla.org/show_bug.cgi?id=1455666

PR #148, opened 2018-05-17
expire access after 90 days note: this is a new variable that is currently only used in dev auth0

gdestuynder:
time in seconds


gdestuynder:
This is live in dev

gdestuynder:
@andrewkrug done.
Note, in restrospect, i think we would make this var simply `expire_access_sec` or similar. Is this what you mean by 'terse variable' on Slack?

gdestuynder:
@andrewkrug

PR #151, opened 2018-05-31
Update apps.yml

jdow:


jdow:
The URL for France Egencia is wrong. Needs updated

apps.yml
```diff
@@ -240,7 +240,7 @@ apps:
   - application:
       name: "Egencia (FR)"
       op: auth0
-      url: "http://www.egencia.fr/pub/agent.dll?qscr=vain&amp;vain=mozilla"
+      url: " http://mozilla.egencia.fr"
```
tristanweir: Should there be a space between the quote and the http?

jdow: Good catch. I recreated the PR #152

PR #167, opened 2018-07-11
update production from master

jdow:


apps.yml
```diff
@@ -826,31 +826,11 @@ apps:
...
-      display: true
-  - application:
-      name: "Mozilla Defense Platform QA"
-      client_id: "slS7eNQ2XUm6yyfB4OPWwprZ7P9OaLMU"
-      op: auth0
-      url: "https://mozdefqa2.private.scl3.mozilla.com/"
-      logo: "mozdef.png"
-      authorized_users: []
-      authorized_groups: ["hris_costcenter_1420", "vpn_opsec_mozdef"]
-      expire_access_when_unused_after: 7776000
```
gdestuynder: what happend here?

gdestuynder: ` expire_access_when_unused_after: 7776000` needs to be ported

PR #182, opened 2018-10-05
Set AAI=LOW for Mozillians

viorelaioia:


gdestuynder:
we should set this only for staging host at this point as production hosts will have to be evaluated before the rollout (im sure mozillians.org will get AAI LOW, but we should use the same process for all RPs)

viorelaioia:
Updated. Can you take another look?

gdestuynder:
reopening with signed commits in #185

PR #198, opened 2018-11-14
change access for splunk as per @johnb

gdestuynder:


gdestuynder:
had johnb check on slack

PR #200, opened 2018-11-19
adding read-only group access

danielhartnell:
I want to provide Hidde with read-only access. In addition to our changes in Auth0, I am adding the read-only group to the apps.yml.


gdestuynder:
@hidde please reopen if you need this, since there was no ping in so long im closing it

PR #220, opened 2019-02-06
new apps

gdestuynder:
Deploys #215 to production and adds new apps to apps.yml


gdestuynder:
thx for catching that

PR #229, opened 2019-02-27
Changing ParSys Jenkins to MEDIUM

hmitsch:
We want 2FA on the ParSys Jenkins server.


hmitsch:
@johngian can you please confirm that you saw this. I want to make sure you know I filed the PR. After that review by kang/andrew would be great.

PR #245, opened 2019-03-26
Add balrog

mozbhearsum:
This is for Balrog prod, specifically. I don't think we need to add dev/stage (correct me if I'm wrong).

I wasn't sure what to do about AAL, but I see lots of apps have it blank, so I did that.

The authorized_groups should be in line with what we do for the current HTTP Basic auth (https://github.com/mozilla-services/cloudops-deployment/blob/master/projects/balrog/puppet/modules/balrog/templates/admin.openresty.conf.erb).


mozbhearsum:
@gdestuynder and/or @jdow - you should probably have a look at this?

gdestuynder:
blank AAL is MEDIUM which is fine :)

PR #265, opened 2019-05-06
opening up gearstore for launch at 4pm today

jdow:


gdestuynder:
note maybe you want hris_is_staff

PR #270, opened 2019-05-20
Fix logos

globau:
- change Heroku logo from Auth0's to Heroku's
- update Slack logo as per new branding
- darken TreeHerder logo to make it readable


globau:
@jdow r?

PR #275, opened 2019-05-22
adding missing additional PTO app

gdestuynder:


gene1wood:
If anyone is curious, MoFo doesn't use the PTO app (hence the grant of rights to only MoCo)

The Mofo vacation policy is outlined in https://mana.mozilla.org/wiki/pages/viewpage.action?pageId=38554955 and use a combination of a Mofo PTO shared calendar and a google form

PR #277, opened 2019-05-24
Add ACLs at Stage 1 access control for more applications, just in case

gdestuynder:


apps.yml
```diff
@@ -2297,3 +2297,1098 @@ apps:
...
+- application:
+    authorized_groups:
+    - hris_is_staff
+    - team_moco
+    - team_mofo
+    authorized_users: []
+    client_id: 5Ju23B8DLiwzd4bzEHOG2fQHe0R4FRwY
+    display: 'false'
+    logo: auth0.png
+    name: Egencia (US)
```
jdow: Egencia is strange is that there are multiple "applications" that all use the same client ID, but have specific IDP-initiated URLs, which Egencia uses to detect where to send the user after login. I'd suggest changing the name of this particular entry to something like "Egencia SAML backend" to indicate that this entry is for access control, whereas the other existing entries are for dashboard display and user routing based on groups.

apps.yml
```diff
@@ -2297,3 +2297,1098 @@ apps:
...
+    display: 'false'
+    logo: auth0.png
+    name: screenshots-admin-default.stage.mozaws.net
+    op: auth0
+    url: https://screenshots-admin-default.stage.mozaws.net/openid/callback/login/
+- application:
+    authorized_groups:
+    - hris_is_staff
+    - team_moco
+    - team_mofo
```
jasonthomas: Please also add 

```
team_mozillaonline
mozilliansorg_stmo_nda
```

apps.yml
```diff
@@ -2297,3 +2297,1098 @@ apps:
...
+    authorized_users: []
+    client_id: VEVkmAppSLa3zc9VRhjkDJPwNSs6Zy7B
+    display: 'false'
+    logo: auth0.png
+    name: screenshots-admin-default.stage.mozaws.net
+    op: auth0
+    url: https://screenshots-admin-default.stage.mozaws.net/openid/callback/login/
+- application:
+    authorized_groups:
+    - hris_is_staff
```
gdestuynder:
```suggestion
    - team_mozillaonline
    - mozilliansorg_stmo_nda
    - hris_is_staff
```

As per Jason

apps.yml
gdestuynder:
```suggestion
    name: Egencia (US) SAML backend
```

PR #278, opened 2019-05-24
string to bool conversion fix

gdestuynder:


vladikoff:
I was there to witness this in my SSO

PR #288, opened 2019-06-14
Bug 1558957: adjust shipit groups

rail:
* Removed unused signoff.shipit.staging.mozilla-releng.net
* Added the `vpn_cloudops_shipit` LDAP group to
  `application.authorized_groups` for shipit (prod and staging). Ship-it
  performs additional checks in order to use refined permissions.
  Additionally the API backend allows only IP addresses from the
  `vpn_cloudops_shipit` VPN IP addresses.

### DIFF
--- a/apps.yml
+++ b/apps.yml
@@ -2348,18 +2348,6 @@ apps:
     - team_moco
     - team_mofo
     authorized_users: []
-    client_id: 5UpSojG2Z4qFxnVS8oa9s3m9KQFpMI5h
-    display: false
-    logo: auth0.png
-    name: signoff.shipit.staging.mozilla-releng.net
-    op: auth0
-    url: https://signoff.shipit.staging.mozilla-releng.net/oidc_callback
-- application:
-    authorized_groups:
-    - hris_is_staff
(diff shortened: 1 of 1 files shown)


rail:
@gdestuynder  can you look at this please

gdestuynder:
@rail does that mean we can delete the client id from our records entirely? (ie will no longer work)

rail:
Yes, we can kill it completely.

wsmwk:
Thanks for moving this forward.  The importance increases as we get into the lower staffed summer months and a busy release period.

gdestuynder:
@rail Please note that `5UpSojG2Z4qFxnVS8oa9s3m9KQFpMI5h` was removed entirely and this client id will no longer work (as per comments above)

@jdow when client ids are removed from this file we should delete them, since this could lead to configuration drift, until automated

rail:
> @rail Please note that `5UpSojG2Z4qFxnVS8oa9s3m9KQFpMI5h` was removed entirely and this client id will no longer work (as per comments above)

:+1: Thank you!

PR #292, opened 2019-07-02
add mozillaonline support to people.mozilla.org

fiji-flo:


gdestuynder:
@jdow any idea why this is needed when there is the `hris_is_staff` bit set? hr issue?

fiji-flo:
looks like it :/

PR #304, opened 2019-07-18
Add NR logo to it-sre team members

The-smooth-operator:
I'd like to have a New Relic icon showing in the dashboard of ITSRE team members. However the application should remain open to team_moco as it is now.

Reading how other applications are configured, this should make it. Could you please check it?


The-smooth-operator:
@gdestuynder thanks for reviewing this. I know there's already an entry for this, I've committed it myself.

Let me explain the details of what I want: 
Currently the app with client id `tAtVU4uyJhaXdMEglSWKxMHliBm9yYtS` is open for `team_moco`, and I want it to remain like this. However I'd like to add an icon to the dashboard for IT SRE team members to access NR. 
If I add the icon in the existing configuration as you suggest, the icon would appear for all the members of `team_moco` and (most of the people) when they click, they won't be able to see anything because they don't have an account on our NR subaccount.

To achieve the behavior described, I copied the structure used by "Netlify" (See entries in lines 2 and 3547).

gdestuynder:
@jdow ive not done this before, can you review it?

jdow:
I think the logic is correct. As long as this new entry doesn't have a client ID, then it should just affect the dashboard itself and do the intended behavior of providing an icon to one set of users, while the other entry handles the access control.

PR #307, opened 2019-07-23
fixing up uptycs config

jdow:


jdow:
> should it be jira_service_accounts?
> otherwise np

Nah, the service_jira group is an ldap group that matches a convention like service_smartsheet, service_uptycs, etc. i.e. "group of people to access a service". The moc_service_accounts is a legacy group of non-human accounts used by monitoring services. Not sure if it's still in use. The new addition is the mozilliansorg_jira_vendors group, whic is a mozillians group for some select vendor people that don't have LDAP accounts.

PR #310, opened 2019-08-13
Update text and link for PMO app

viorelaioia:


viorelaioia:
@gdestuynder , @fiji-flo , can you please take a look over these changes? Thanks!

PR #316, opened 2019-08-21
Add k8s discourse dev to apps.yml

LeoMcA:
Discourse needs this to get an AAL value in id_tokens


LeoMcA:
Whoops, client id is already there. I didn't check first

PR #318, opened 2019-08-30
adding Console View acl

jdow:


jdow:
https://bugzilla.mozilla.org/show_bug.cgi?id=1576983 is all I have for info

jdow:
Based on the domain there, maybe it should be called ConsoleFlow ?

apps.yml
```diff
@@ -3755,3 +3755,14 @@ apps:
...
     url: https://fxa-support-panel.prod.mozaws.net/openid/callback/login
+- application:
+    authorized_groups:
+    - team_netops
+    - team_infra
+    authorized_users: []
+    client_id: OSKMKPkvUoDg0K8agd4iHSwUam0lpr6p
+    display: false
+    logo: auth0.png
+    name: Console View
```
gene1wood:
```suggestion
    name: ConsoleView
```

jdow: I'm switching it from View to Flow, since that's the domain

gene1wood: Oh right, good catch.

PR #319, opened 2019-09-12
adding AAL: Low to IM test apps

jdow:


mhoye:
I confirm gdestuynder's assertion, we can go ahead with this.

satdav:
can we show this now on the dashboard @mhoye

PR #331, opened 2019-10-26
Update apps.yml

gdestuynder:
Remove community analytics


gdestuynder:
Site no longer exist/verified (https://analytics.mozilla.community/)
@jdow this should also be removed from auth0 ideally

jdow:
I've deleted it in auth0.

PR #332, opened 2019-10-28
Remove HackMD

gene1wood:
This pilot has been torn down so we can remove it from this side as well


gene1wood:
I've requested removal of the RP from Auth0 in [Bugzilla 1591941](https://bugzilla.mozilla.org/show_bug.cgi?id=1591941)

PR #333, opened 2019-10-30
apps: remove `IRCCloud` as an app on the dashboard

cgsheeh:
The Mozilla IRCCloud instance has been decommissioned
for just under a month now. Time to remove it from the
dashboard.


cgsheeh:
See https://github.com/mozilla-iam/sso-dashboard-configuration/pull/335 instead.

PR #339, opened 2019-11-13
Add New Relic EIS

gene1wood:


apps.yml
```diff
@@ -3788,3 +3788,14 @@ apps:
     name: Cloudsnap
     op: auth0
     url: https://auth.mozilla.auth0.com/samlp/cuC3NOGPFm58H172NnOypPf1jGQwMoUf
+- application:
+      authorized_groups:
+        - team_opsec
```
gdestuynder: when i login it looks like ive access to non-eis accounts (just like pre-sso)

Shouldnt this be available to more than just `team_opsec`?

gdestuynder: hum "actually" once i logged in with the old account it merged them for this

gdestuynder: Probably better
```suggestion
        - hris_costcenter_1420
```

gene1wood: Yes, there shouldn't be any constraint from the auth0 side. Do I just remove the `authorized_groups` or just leave it as an empty list?

gene1wood: I've changed this to `team_moco`

PR #345, opened 2019-11-19
Reflect New Relic RPs.

The-smooth-operator:
The reason why not displaying the icon is because AFAIK the members of each NR account are not matching with any pre-existent group


jdow:
These entries don't have a client_id which is what is used for access control. Without a client ID, an entry is only useful for dashboard display, but since that is false, I think this is not needed.

The-smooth-operator:
@jdow should I add the client_id field for each entry?

jdow:
Yes, which should be the string in the URL. For  https://auth.mozilla.auth0.com/samlp/Um2rLocE3s851JXNZzTPnA5DFzWe9OhQ , the client ID is Um2rLocE3s851JXNZzTPnA5DFzWe9OhQ

The-smooth-operator:
thanks, I've updates the PR.

gdestuynder:
@jdow

PR #351, opened 2019-12-12
adding apps.yml entry for consider.it

jdow:


apps.yml
```diff
@@ -3886,4 +3886,15 @@ apps:
...
+    AAL: LOW
+    authorized_groups:
+    - everyone
+    authorized_users: []
+    client_id: p2lSKumiDFFnTfopf6zkiF3WB0W359ZK
+    display: true
+    logo: auth0
+    name: Consider.it
+    op: auth0
+    url: https://mozilla.consider.it/
```
tkriplean: Is this the correct url? I believe we're using https://mozilla-soti.consider.it

tkriplean: Hi @jdow and @gdestuynder, sorry to be annoying, but could we address this pull request? Working on a tight deadline (monday 12/13) unfortunately.

jdow: I fixed the URL.

apps.yml
```diff
@@ -3886,4 +3886,15 @@ apps:
     op: auth0
     url: https://auth.mozilla.auth0.com/samlp/6RMuGl43VwCGOxvFXoDLDXE3U3yhEtT8
     vanity_url:
-    - /securitytraining
\ No newline at end of file
+    - /securitytraining
+- application:
+    AAL: LOW
```
gdestuynder: is the service owner aware of what LOW means for their site?

tkriplean: Hi @gdestuynder, I believe I do understand. Correct me if I'm wrong, but the LOW AAL refers to low risk as described at https://infosec.mozilla.org/guidelines/risk/standard_levels. I believe that low risk is accurate as documented there. The harm from a compromised account accessing considerit and adding their opinions is very low.

At a high level, I don't want to force accounts to have 2FA enabled to access the forum. It is not the default for any of the account modalities supported via Mozilla's auth0 portal, and the error messages delivered to users without 2FA enabled are confusing (it states that Considerit is the one requiring "higher security", and suggest using a Firefox account, even though Firefox accounts also don't have 2FA enabled by default...). So if there is a better user experience AND higher security, I'm good with that too.

tkriplean: In case it is instructive, this is the error message I receive after trying to authenticate via auth0 with a brand new Firefox account: 
Screen Shot 2019-12-11 at 4 57 56 PM

gdestuynder: The default for all reliers is "MEDIUM" which requires MFA. Firefox Account with MFA setup would work for example (but it *does* require MFA to be setup). Also, "Staff" logins always require MFA regardless of this.

Setting "LOW" does not require MFA, which also means accounts are more likely to be compromised.

Notes:
MFA is not the only check, but it is one of the main difference between LOW and MEDIUM. We recommend using MFA for most services unless it impacts community adoption for example. 

Even when that's the case, we often set it up that way:
- if user logged in as "LOW" assurance, they can only do basic tasks (forum chat, voice contribution on common voice, etc.)
- if user logged in as "MEDIUM" or higher assurance, they can do more sensitive tasks (admin type tasks, forum curation, etc.).
- certain reliers will differentiate further (such as "HIGH" for admin actions) but that's discretionary and we consider "MEDIUM" to be otherwise sufficient for all tasks.

With that said, if your relier only provides the "basic tasks" scenario, then "LOW" and you need a higher adoption rate by allowing non-MFA'd accounts/accounts that may fail other security assurance checks, then setting to "LOW" indeed makes sense. If that's not the case, we should probably chat on Zoom before you accept this kind of risk

Lastly: we could definitely improve the error message and the handling. If simply improving the error message makes using "MEDIUM" acceptable for you, we can definitely do this very quickly.

Improving the flow itself is on our backlog but will take much longer.

tkriplean: LOW is fine for this application. Folks can really only use their accounts for adding opinions and discussing. Only a small number of people will have admin privileges, and aside from me, they're all mozilla staff as far as I know. So they should have 2FA set up. 

Which is a good reminder that I need to setup 2FA on those new email/Firefox accounts I registered to test the system!

I think y'all should improve the error messaging, but setting to LOW here is best IMHO. (and if in the future you want external feedback on improved error messaging, feel free to ping me)

nukeador:
As requested, I comment here to acknowledged we want to allow non-2fa logins on community.mozilla.org to enable community adoption. User iterations should be considered LOW risk due their nature on this site.

Thanks!

PR #356, opened 2020-01-15
remove akrug, jmize (jgmize). Add aalexander to "GCP Infrastructure" RP

metadave:


gdestuynder:
yes its probably better. we used the access file for people where we wanted to 100% ensure access evenif theres an issue with mozillians.org

PR #357, opened 2020-01-16
Clarify the logic around authorized_users and authorized_groups

gene1wood:


README.md
```diff
@@ -39,8 +39,12 @@ This is a list of available fields.
       vanity_url: ['/an-easy-to-remember-url']
 
       ### Security settings
-      # The list of users and groups allowed to access this RP. Both empty means allow everyone, one empty means the
-      # other field is used
+      # The list of users and groups allowed to access this RP. 
+      # If both authorize_users and authorized_groups are empty, everyone is allowed
+      # If one is empty and the other has content, only the members of the non empty one are allowed
+      # If both have content, the union of everyone in both are allowed
+      # https://github.com/mozilla-iam/auth0-deploy/blob/4ac5cb4959fc93a668fcc9909ce33eac2eb8416c/rules/AccessRules.js#L173-L181
```
gdestuynder: i would be careful pointing to specific commit as if the code change these will still point to the older version. you might want to point to "whatever master or production is at"

gene1wood: The reason for the specific commit is that when you anchor to line numbers, you should use a commit as otherwise the content of those line numbers will change over time.

PR #370, opened 2020-03-27
adding getabstract and udemy

jdow:


april:
Should getabstract test use the getabstract logo instead of the auth0 logo? Or was that intentional?

jdow:
getAbstract test is keeping display set to false, so no logo is required

PR #371, opened 2020-04-10
Add missing client_ids

gene1wood:
These RPs don't have client_id fields so I've added them.


gene1wood:
@jdow Thanks for taking a look

> The Egencia one because all the different Egencia apps all use a single SAML configuration, but different URLs based on a user's location data.

Indeed. I've checked the Auth0 rules and confirmed that the impact of having multiple entries in `apps.yml` with the same client_id only comes into play if the record also uses the `expire_access_when_unused_after` field, which none of these do. I wrote up some comments about this [in this PR](https://github.com/mozilla-iam/auth0-deploy/pull/320/files#diff-82b33567465094c8d23f4dc452cc798fR135-R140).

In looking through the sso-dashboard codebase, I don't see it using the `client_id` value

Are there other impacts that I'm not thinking of, of having multiple entries in `apps.yml` with the same client_id

> Egencia has it's own access control, managed by Workday and WPR.

Ya, that was my understanding.

> The SpaceIQ preview is also without client ID, as it's a duplicate with another entry that does have a client ID. In this case, we want to allow all staff to access it, but only a select preview group to have the dashboard icon.

Would adding the `client_id` to this affect the access? Is there something I'm overlooking?

> I think for Expensify, we might be able to add the client ID, but I'm unsure if it will break a workflow for someone somewhere. It also has its own access control managed by Finance/Accounting team that I have no visibility into (i.e. I'm unsure if there are non-SSO or non-staff SSO users in Expensify).

Ya, I don't think that the client_id affects access.

jdow:
I think the problem with Egencia is that the sso dashboard code uses the authorized_groups in conjunction with the display true/false field to show the icon to specified groups. The Auth0 access control rule uses the authorized_groups in conjunction with the client ID to manage access control. The way these are set up right now, they are not doing any access control because of the lack of client ID, and they each have a separate set of groups associated with them, for display purposes. If we were to add a client ID to any of them, then all of a sudden, the authorized_groups would come into play with access control, and since there are multiple apps, with the same client ID, I assume that the first one encountered would win, so assuming it's top-down, then adding a client ID, would restrict the Egencia app to only people in the US and Canada groups, blocking everyone else. I don't see a good way to fix that, while retaining the per-group dashboard settings that are currently in place. It's a similar story with the Space IQ apps being doubled up, one with display true, and the other with a client ID. An ultimate fix would be to rewrite portions of the dashboard and the auth0 access control rule to stop overloading the use of the authorized_groups field, and have instead distinct "authorized_access_groups" and "display_to_groups".

gene1wood:
This PR now only adds missing client_ids for New Relic and Expensify

ctbfourone (approved):
+1, verified client IDs, looks good

PR #379, opened 2020-05-14
adding fxa-admin-panel prod app

jdow:


jdow:
I've updated the name in auth0 and in this PR

PR #380, opened 2020-05-14
Remove the mozilla-iam AWS account SAML integration client

gene1wood:
See IAM-543


jdow:
Should I also remove the ﻿aws_320464205386_admin LDAP group, or is it used elsewhere? @gene1wood

gene1wood:
@jdow No the `﻿aws_320464205386_admin` LDAP group is used elsewhere.

PR #392, opened 2020-08-27
Replace moderator staging with updated one

pwnbus:


gene1wood:
This is deployed to production today.

PR #393, opened 2020-09-17
Open SSO for AWS up to everyone

gene1wood:
This removes the staff only restriction for AWS SSO


gene1wood:
I think we'll instead leave it closed to staff and just add in explicit access groups which contain non-staff instead.

PR #403, opened 2020-11-03
Add vouches-mozillians application

The-smooth-operator:


apps.yml
```diff
@@ -571,6 +571,16 @@ apps:
...
     vanity_url:
     - /mozillians
+- application:
+    AAL: LOW
+    authorized_groups:
+    - everyone
+    authorized_users: []
+    client_id: HC1FeMf3dVCTnAZbQR08quMylQEUcu60
+    display: false
+    name: vouches-mozillians
```
gene1wood, line 581: This should be a human readable name like `vouches.mozillians.org` or `Mozillians Vouches` or something
```suggestion
    name: vouches.mozillians.org
```

apps.yml
```diff
@@ -571,6 +571,16 @@ apps:
...
+    op: auth0
+    url: https://vouches.mozillians.org/oidc/callback
```
gene1wood, line 583: Should this be `https://vouches.mozillians.org/oidc/authenticate/` as that's the login URL for the site (as opposed to the callback URL)?
```suggestion
    url: https://vouches.mozillians.org/oidc/authenticate/
```

gene1wood (changes requested):

apps.yml
The-smooth-operator: I'm a bit confused on what the `url` field should be here. Many of the current configurations are listing a callback URL while others are listing the main URL of the site.
Based on what you are saying, should it be always the login URL of the site?

The-smooth-operator: according to the documentation:
```
      ### SSO Dashboard display settings
      # This is the canonical URL for the RP / front page
      url: "https://rp.example.net/"
```
So maybe the correct value should be: `url: https://vouches.mozillians.org/` ?

gene1wood: So the `url` value is the link that's used on the SSO dashboard page. In this case since we're not showing it on the dashboard it's not super important but we should set it to the right value (in case we ever turn it on for the dashboard).

So the goal should be for the URL to be what the user hits that gets them onto the target site in a logged in state.

So for something like discourse, if you set the `url` to https://discourse.mozilla.org/ then the user gets taken to the main page of discourse but they're not logged in, they'd have to click the login button. Instead, the `url` is set to https://discourse.mozilla.org/auth/auth0 which is the URL that a user hits when the click the login button.

For any site that has their callback URL set as the `url` value, this is probably wrong. It may work since if it's a site where all pages are protected by SSO (unlike discourse or people.mozilla.org for example), then when an unauthenticated user hits the callback URL, the site just sees that they're unauthenticated and sends them to the correct login URL. It's not ideal as it's an extra unnecessary redirect. If it's a site like vouches or discourse, it won't work.

Since vouches.mozillians.org works like discourse and people.mozilla.org, in that not every page is protected by SSO (for example the homepage). If you browse to the callback URL ( https://vouches.mozillians.org/oidc/callback ) unauthenticated, it just redirects you to the main page of the site and you remain unauthenticated (which isn't ideal).

So the right thing to do here is to set it to the URL that when the user goes to it, they get logged in, which in this case is https://vouches.mozillians.org/oidc/authenticate/

I'll PR that documentation page to make this clear.

The-smooth-operator: Completely understood, thank you! I've updated the PR.

gene1wood (approved):

PR #408, opened 2020-12-07
Display Matrix tile in SSO dashboard

The-smooth-operator:


limed (approved):

gene1wood (approved):

The-smooth-operator:
The change has been deployed on dev, and I can verify it's working.
I found an issue with the search bar not behaving as expected, will investigate more an file a ticket for it.

PR #409, opened 2020-12-08
Create TripActions application

The-smooth-operator:
This PR will enable SSO for TripActions application. See [Bugzilla](https://bugzilla.mozilla.org/show_bug.cgi?id=1672012) for more info

### DIFF
--- a/apps.yml
+++ b/apps.yml
@@ -4056,3 +4056,13 @@ apps:
     name: Smartling
     op: auth0
     url: https://sso.smartling.com/sso-apps/dashboard/accounts/3f16c2f8
+- application:
+    authorized_groups:
+    - mozilliansorg_tripactions
+    authorized_users: []
+    client_id: i1qBrMjJEdKlTNEVgGwc9P0xFxI9cuD8
+    display: false
+    logo: auth0.png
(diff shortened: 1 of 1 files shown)


limed (approved):
r+

gene1wood (changes requested):
Alberto, much better than granting individual users access to an RP is to identify a curator who can curate an access group membership, and add that access group. That way in the future if there are changes to membership, it need not require you to make code changes in this repo, and the group curator can instead just add or remove users from the group. Users can create access groups here https://people.mozilla.org/a/ and if they do you can then add them to the `authorized_groups` field.

For example if the curator creates group `foo` then you'd add `mozilliansorg_foo` to the `authorized_groups`

The-smooth-operator:
Gene, this makes complete sense. I've added a new group and pushed (squashed) the changes

gene1wood (approved):

gene1wood:
Also it looks like the URL is actually `https://auth.mozilla.auth0.com/samlp/i1qBrMjJEdKlTNEVgGwc9P0xFxI9cuD8`. I'll put in a PR to fix that.

PR #410, opened 2020-12-08
Create app for buzilla-dev.allizom.org

The-smooth-operator:
See also https://bugzilla.mozilla.org/show_bug.cgi?id=1671709

### DIFF
--- a/apps.yml
+++ b/apps.yml
@@ -4045,4 +4045,16 @@ apps:
     name: experimenter.stage.mozaws.net
     op: auth0
     url: https://stage.experimenter.nonprod.dataops.mozgcp.net/
+- application:
+    authorized_groups:
+    - hris_is_staff
+    - team_moco
+    - team_mofo
+    authorized_users: []
+    client_id: K1fsiDh5Ta7pc7BrEECi2VLhPyqHwy4r
+    display: false
+    logo: auth0.png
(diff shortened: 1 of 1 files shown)


gene1wood (commented):
Is the `url` value in this PR correct?

https://github.com/mozilla-iam/sso-dashboard-configuration#fields-reference
```

      ### SSO Dashboard display settings
      # This is the URL that a user must visit to be logged into the RP. This URL would
      # either be the URL of the login button on the site (if it has one), or the URL
      # that a user gets redirected to when they visit a protected page while unauthenticated.
```

The-smooth-operator:
> Is the `url` value in this PR correct?

Thanks for your review @gene1wood. 
I did take into account that the URL should be the target where a user can login in.
In this case, the login URL is the same as the main URL because when we click "Log in" we get prompted with an embedded form and not getting redirected to other login page.

image

It's also true that the app rewrites the URI to to `/home` but since there aren't any redirects (no more hops) I thought this will be more stable.

limed (approved):
r+ there is a conflict so you have to resolve that first

The-smooth-operator:
Once the integration is setup, and if we want to display the URL in the dashboard for prod we will pick the URL which brings the user to the site authenticated

PR #414, opened 2020-12-17
Adding tripactions tile

limed:


limed:
Requesting bug is https://bugzilla.mozilla.org/show_bug.cgi?id=1672012

This adds the tripactions tile to the sso dashboard

apps.yml
```diff
@@ -4030,7 +4030,7 @@ apps:
     authorized_users: []
     client_id: i1qBrMjJEdKlTNEVgGwc9P0xFxI9cuD8
     display: false
```
The-smooth-operator, line 4032: You have to set `display: true` for the tile to be displayed

The-smooth-operator (changes requested):
Thanks for doing this Ed!

apps.yml
limed: Fixed in ed01bf844006be2bf409692382fea31d774018ad

The-smooth-operator (approved):
looks good now, thanks!

PR #415, opened 2020-12-17
Remove all instances of developer portal

limed:
Developer Portal has been decommed so we no longer need these


gene1wood:
@limed have you deleted these client IDs from the prod Auth0 tenant?

limed:
> @limed have you deleted these client IDs from the prod Auth0 tenant?

Yes I have its gone

The-smooth-operator:
For some reason it doesn't allow to add me as a reviewer. Merging this

PR #418, opened 2021-01-06
Cleanup references to old phonebook deployments

The-smooth-operator:
This PR will delete all the occurrences of Phonebook in the configuration.
I've also deleted the respective apps in Auth0 for all applications but `o2e391VjmnPk0115UedNTmRL8x2nySOa` as I'm not sure if that can go.
Please @fiji-flo can you confirm you are not using that one but `o2e391VjmnPk0115UedNTmRL8x2nySOa`?

### DIFF
--- a/apps.yml
+++ b/apps.yml
@@ -1704,6 +1704,8 @@ apps:
     name: Mozilla People Directory
     op: auth0
     url: https://people.mozilla.org
+    vanity_url:
+    - /phonebook
 - application:
     AAL: LOW
     authorized_groups:
--- a/apps.yml
+++ b/apps.yml
@@ -643,28 +643,6 @@ apps:
     - /pagerduty
 - application:
     authorized_groups:
-    - phonebook_access
-    authorized_users: []
-    client_id: K7vKewjQHKe45mmOo5cRae6yyOvnmg74
-    display: false
-    logo: phonebook.png
-    name: Phonebook
-    op: auth0
-    url: https://auth.mozilla.auth0.com/samlp/K7vKewjQHKe45mmOo5cRae6yyOvnmg74
-- application:
-    authorized_groups:
-    - phonebook_access
-    authorized_users: []
-    client_id: 00fgOKsjo530sIxfhsved8jyTjAD0av2
-    display: false
-    logo: phonebook.png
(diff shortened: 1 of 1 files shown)


fiji-flo (approved):
Looks good. `o2e391VjmnPk0115UedNTmRL8x2nySOa` is used by people.mozilla.org (the second appearance is just to make sso.mozilla.com/phonebook work, and we might wanna keep it).
Please make sure to also delete all credentials in auth0 as `apps.yml` is also used to assert access groups.

apps.yml
```diff
@@ -1726,19 +1704,6 @@ apps:
...
-    AAL: LOW
-    authorized_groups:
-    - everyone
-    authorized_users: []
-    client_id: o2e391VjmnPk0115UedNTmRL8x2nySOa
-    display: true
-    logo: people.png
-    name: Phonebook
-    op: auth0
-    url: https://people.mozilla.org
```
limed, line 1739: This particular removal references people.mozilla.org is this removal correct?

The-smooth-operator:
> Looks good. `o2e391VjmnPk0115UedNTmRL8x2nySOa` is used by people.mozilla.org (the second appearance is just to make sso.mozilla.com/phonebook work, and we might wanna keep it).
> Please make sure to also delete all credentials in auth0 as `apps.yml` is also used to assert access groups.

Ok, thanks for checking this.

apps.yml
The-smooth-operator: This is kinda correct. As Florian says above, that client ID is duplicated, because it is used also in [here](https://github.com/mozilla-iam/sso-dashboard-configuration/blob/master/apps.yml#L1743-L1753).

The-smooth-operator:
So, proceeding with the deletion of the app commented by you too will break the link sso.mozilla.com/phonebook 
What I'm going to do is to add that vanity URL to the current Mozilla People Directory app. So we achieve best of both world: we don't break people's bookmarks and we clean up.

The-smooth-operator:
done. So I won't delete the app in auth0 because it is still in use.
If it's looking good I'll squash the commits before merging.

fiji-flo (approved):

limed (approved):

PR #437, opened 2021-02-23
Fix tripactions URL

gene1wood:
The existing URL doesn't log the user in. Replacing it with the login URL
Updates #409


duallain (approved):

duallain:
It doesn't log me in (as I don't believe I have access to trip actions) but the error message I get seems reasonable.

PR #438, opened 2021-02-23
Update access groups for netlify and newrelic

limed:


duallain (approved):
Seems good, generally just moving from ldap access to people.mozilla.org groups, 👍

PR #439, opened 2021-02-24
Adding tile for marketing's email service Everest

duallain:
Service provided by validity.
Related to the setup request from https://bugzilla.mozilla.org/show_bug.cgi?id=1686120


duallain:
@gene1wood this is the first one of these I've done, so a bit more scrutiny is probably warranted, but I took a look at a few previous PRs and this seems to be good based on those.

gene1wood (commented):
Alan, this looks correct. Group looks good. URL redirects to SSO login which is correct. It looks like the `everest.png` is an "empty file"? Is that expected? I'd assume it would have some size as it would be the tile image.

duallain:
@gene1wood not sure honestly, I see what you're talking about with the file being listed as empty, but on both my machine and hitting it directly (like the below link) it seems to look alright.

(I just checked again and now it's showing binary file data not shown, which is the correct behavior.  Happened after in the pr/files view I selected the 'display rich preview' button and then back to 'display source diff')

https://github.com/mozilla-iam/sso-dashboard-configuration/blob/dd2432851c43b84cba663a5ea7af78998131903c/images/everest.png

gene1wood (approved):
Got it, ok just some weird GitHub UI issue.

PR #435, opened 2021-02-22
Grant MozillaOnline employees access to telemetry services

The-smooth-operator:

PR #441, opened 2021-02-26
Grant MozillaOnline employees access to telemetry services (cont'd)

l-hedgehog:
As a followup to #435, which doesn't include the other three telemetry related services as confirmed by @jasonthomas in https://bugzil.la/1693956#c4 .

I'm not sure which one of "GLAM" or "GLAM (prod)" is the right one, so updated them both.

r?@The-smooth-operator

Thanks!

### DIFF
--- a/apps.yml
+++ b/apps.yml
@@ -2147,6 +2147,7 @@ apps:
     - hris_is_staff
     - team_moco
     - team_mofo
+    - team_mozillaonline
     authorized_users: []
     client_id: OlwWsXslbA9wk5lQOHUDIUSrtIFTauTy
     display: false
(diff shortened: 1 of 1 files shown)


sciurus:
@l-hedgehog the-smooth-operator is out on leave.

LTGM. @duallain can you provide a second set of eyes?

duallain (approved):
LGTM.   Based on the previous conversation we OK'd this access, so it makes sense to show the tiles on the dashboard.

sciurus:
@l-hedgehog can you sign this commit?

https://docs.github.com/articles/about-gpg/

l-hedgehog:
> @l-hedgehog can you sign this commit?

Hi @sciurus I pushed a signed commit.

PR #442, opened 2021-02-26
mentoring app is on dotcom not dotorg

sciurus:
https://jira.mozilla.com/browse/SE-1483

### DIFF
--- a/apps.yml
+++ b/apps.yml
@@ -668,7 +668,7 @@ apps:
     logo: auth0.png
     name: Mentoring
     op: auth0
-    url: https://mentoring.mozilla.org/oidc/authenticate/
+    url: https://mentoring.mozilla.com/oidc/authenticate/
     vanity_url:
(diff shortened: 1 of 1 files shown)


gene1wood (approved):

duallain (approved):
LGTM.  One note @sciurus we may need to update the auth0 client to allow .com access, it seems like maybe the callback url isn't currently allowed.

gene1wood:
> we may need to update the auth0 client to allow .com access, it seems like maybe the callback url isn't currently allowed.

Good point, if the URL of the allowed callbacks is also wrong in Auth0 it will need to be updated.

sciurus:
@gene1wood @duallain when I check allowed callback URLs in Auth0 I see "https://mentoring.mozilla.com/oidc/callback/" which looks right.

sciurus:
Force-pushing a signed version of the commit

PR #443, opened 2021-03-03
Update snippets url

duallain:


gene1wood (approved):
If this changes the callback URL you'll want to update that in Auth0

PR #446, opened 2021-03-16
Remove team_pocket from apps

gene1wood:
Given that all members of team_pocket are now members of team_moco and the purpose of
the team_pocket group has changed, this commit removes those references to
team_pocket except in cases where it is explicity granting access to only the members

### DIFF
--- a/apps.yml
+++ b/apps.yml
@@ -532,7 +532,6 @@ apps:
     - hris_is_staff
     - team_moco
     - team_mofo
-    - team_pocket
     authorized_users: []
     client_id: 80JNexePA737rSLhBAABqIvMJTEAn11u
     display: false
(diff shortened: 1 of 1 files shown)


rtucker-mozilla:
This still leaves the team_pocket reference at the only place that I can imagine caring (getpocket.atlassian.net) r+ from me!

gene1wood:
`team_pocket` grants access to these 4 sites
* Slack Pocket
* getpocket.atlassian.net
* recit.getpocket.dev
* 250ok

sciurus:
LGTM; I see team_pocket remaining for the 4 sites gene mentioned.

duallain (approved):

PR #447, opened 2021-03-17
Adding proper login url and vanity endpoint for getpocket jira

rtucker-mozilla:
This should improve the getpocket jira sso experience.


rtucker-mozilla:
Ended up having a typo that the build system found, neat!

PR #451, opened 2021-03-22
Adding jira_cloud_access to groups able to access mozilla-hub.atlassi…

rtucker-mozilla:
…an.net, which also solves for sso access into getpocket.atlassian.net

Signed-off-by: Rob Tucker <rtucker@mozilla.com>


sciurus (approved):
I see https://people.mozilla.org/a/jira_cloud_access/ so LGTM

PR #452, opened 2021-03-23
[DO NOT MERGE BEFORE 4/19] Make culture amp available to everyone

sciurus:
This deduplicates the culture amp config and makes it visible to all employees as requested in https://bugzilla.mozilla.org/show_bug.cgi?id=1700222

### DIFF
--- a/apps.yml
+++ b/apps.yml
@@ -3885,6 +3885,7 @@ apps:
     - team_mofo
     - team_mozillaonline
     authorized_users: []
+    client_id: axRGaXPRd5pe60fnEllluSN76MefFX1E
     display: true
     logo: cultureamp.png
     name: Culture Amp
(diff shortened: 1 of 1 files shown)


sciurus:
Requester does not want this to happen until week of 4/19.

apps.yml
```diff
@@ -3885,16 +3885,6 @@ apps:
     - team_mofo
     - team_mozillaonline
     authorized_users: []
-    client_id: axRGaXPRd5pe60fnEllluSN76MefFX1E
```
gcoxmoz, line 3888: If this line drops, the resulting entry has no `client_id`.  Intentional?

sciurus: Thanks for catching that! I had accidentally deleted one more line than I intended.

rtucker-mozilla (approved):
r+

gcoxmoz (approved):
My review is not worth much, but this appears to contain parallelism with other stanzas as to match for other things that are working.

PR #454, opened 2021-03-25
Standardize New Relic Access

sciurus:
Give everyone NR access. Do not display it for anyone, except I think
for Web SRE who gets an override earlier in the file.

This was prompted by  https://mozilla-hub.atlassian.net/browse/SE-1522


apps.yml
```diff
@@ -3527,9 +3528,11 @@ apps:
     url: https://auth.mozilla.auth0.com/samlp/cuC3NOGPFm58H172NnOypPf1jGQwMoUf
 - application:
     authorized_groups:
-    - hris_costcenter_1420
```
sciurus, line 3530: I think these hris groups don't exist, so effectively everyone has access now.

Since a user has to be manually created on the New Relic side, this isn't a security issue.

gene1wood: > I think these hris groups don't exist, so effectively everyone has access now.

If no entry exists in apps.yml then everyone has access. 

If an entry exists in apps.yml and contains no groups or users then everyone is denied access

https://github.com/mozilla-iam/auth0-deploy/blob/f8539acfbc803ce2d8fba2c932384db4b0380431/rules/AccessRules.js#L173-L177

If an entry exists in apps.yml and contains either authorized groups or users then users are allowed that are members of those groups or who are those users

https://github.com/mozilla-iam/auth0-deploy/blob/f8539acfbc803ce2d8fba2c932384db4b0380431/rules/AccessRules.js#L180-L188

So the question is, why do I have access. I [can confirm I'm not a member of](https://github.com/mozilla-iam/auth0-deploy/blob/f8539acfbc803ce2d8fba2c932384db4b0380431/rules/AccessRules.js#L117) `hris_costcenter_1420`. Reading through the logic I can't figure out why I have access. I'm going to dig a bit further

gene1wood: Aha, it's a problem I introduced in #339 . I didn't include the `client_id` so as a result, this stanza in the config has never been used.

So we need to add a client_id here for this section to matter at all.

That's why it's been open to everyone, because essentially it's missing from apps.yml

sciurus: Ah,  I forgot it was missing that! I actually add the client ID in this PR.

duallain (approved):

gene1wood (commented):
So there's an additional record that's for this up at line 22 that uses a mozillians group. Maybe simpler would be to consolidate these two stanzas that use the same client_id (one explicit and one that will be present after this PR is merged). Basically, make the same access group control both who has access and who sees it in the sso dashboard tiles.

ctbfourone:
Closing in favor of https://github.com/mozilla-iam/sso-dashboard-configuration/pull/658

PR #455, opened 2021-03-25
Switch basket from ldap to mozillians

sciurus:
https://people.mozilla.org/a/basket_admin/

At pmac's request


sciurus:
Don't deploy until pmac says so.

duallain (approved):

pmac:
I think we're good to go for this change. Thanks!

PR #458, opened 2021-03-26
Switch Snippets and Nucleus to Mozillians group

pmac:
We'll need the `snippets_admin` and `nucleus_admin` groups created on people.m.o first, but I think this should do it.

/cc @sciurus @duallain

### DIFF
--- a/apps.yml
+++ b/apps.yml
@@ -696,7 +696,7 @@ apps:
     url: https://vouches.mozillians.org/oidc/authenticate/
 - application:
     authorized_groups:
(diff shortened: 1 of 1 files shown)


sciurus:
@pmac i created the groups and made you curator.

pmac:
Thanks! I've invited all of the necessary people to the groups.

pmac:
I think we're good to go for this switch. Thanks again for your help!

sciurus (approved):

PR #459, opened 2021-04-05
Adding Planful to SSO dashboard. [1680831]

rtucker-mozilla:
Adding Planful to SSO dashboard. [1680831]


apps.yml
```diff
@@ -4029,3 +4029,15 @@ apps:
...
+- application:
+    authorized_groups:
+    - mozilliansorg_finance_planful
+    authorized_users: []
+    client_id: H5ddlJSCfGP8ab65EnWaB2sd541CJAlM
+    display: true
+    logo: planful.png
+    name: Planful
+    op: auth0
+    url: https://planful.com/server-login/
```
gene1wood, line 4041: You likely want to change this to the login URL as it looks like that URL doesn't start the SSO login process
```suggestion
    url: https://auth.mozilla.auth0.com/samlp/H5ddlJSCfGP8ab65EnWaB2sd541CJAlM
```

Currently the `H5ddlJSCfGP8ab65EnWaB2sd541CJAlM` client in Auth0 has no login methods enabled so this URL won't work yet.

gene1wood (changes requested):

PR #475, opened 2021-06-02
Add Papertrail Web SRE app and image [Bug 1691717]

dividehex:


apps.yml
```diff
@@ -4058,3 +4058,15 @@ apps:
...
+- application:
+    authorized_groups:
+    - mozilliansorg_web-sre-papertrail-access
+    authorized_users: []
+    client_id: QSbhAzqUlqCSWt6iAV45um5DDGdhhDTR
+    display: true
+    logo: papertrail-logo.png
+    name: Papertrail Web SRE
+    op: auth0
+    url: https://auth.mozilla.auth0.com/samlp/QSbhAzqUlqCSWt6iAV45um5DDGdhhDTR
```
duallain, line 4070: This url is a little different than what I would generally expect.  I'd expect it to be the papertrail domain with whatever path gets you to a nice login page.

apps.yml
```diff
@@ -4058,3 +4058,15 @@ apps:
...
+    vanity_url:
+    - /papertrail-web-sre-access
```
floatingatoll, line 4072: I'd request `papertrail-websre` instead here, since "web-sre-access" is hyphenated oddly and doesn't map well onto a 'guessable' short URL here.

apps.yml
gene1wood: This would be a normal URL for a SAML integration. the `url` field should be the URL that when you browse to it, it continues a redirect through to have you logged into the site.

dividehex: Afaik, this URL is consistent with many application already in place.  In my testing, I also discovered, the Solarwinds Unified login page doesn't work if you do not enter the aforementioned URL at least once.  On first authentication, it creates a local user based on role/group mapping from the SAML response.  If you do not have the proper roles in your SAML response, you are bounced back to the Solarwinds Login page.

I'll also note here that each Solarwinds org will need its own auth0 SAML application.  If your email address is assosicated with multiple orgs, like mine is, Solarwinds will SAML authenticate you for each one that has SAML enabled.  If you are also a member of a non-SAML enabled org, you will be instructed to log into Solarwinds with a standard Solarwinds account in order to authenticate you to any remaining orgs.

apps.yml
dividehex: Good suggestion!

duallain (approved):
The url currently takes me to a solarwinds page with an error (SAML configuration not enabled), but everything looks good to me otherwise.  If this follows the pattern of other saml apps then shipit 🚀

floatingatoll (approved):
I think this is fine.

dividehex:
> 
> 
> The url currently takes me to a solarwinds page with an error (SAML configuration not enabled), but everything looks good to me otherwise. If this follows the pattern of other saml apps then shipit 🚀

Yes, the URL will not work until I flip the switch in papertrail, which I did not want to do until the application was in place.  It should work soon. 😸

PR #476, opened 2021-06-02
Add application for Kevel [Bug 1713035]

dividehex:
https://bugzilla.mozilla.org/show_bug.cgi?id=1713035

This PR sets up a new SAML application to supersede the AdZerk application.  Once Kevel applies the SAML metadata to their side and we can verify the SSO works on our side, I will then submit a follow up PR to remove the deprecated AdZerk application.

### DIFF
--- a/apps.yml
+++ b/apps.yml
@@ -3454,6 +3454,16 @@ apps:
     url: https://auth.mozilla.auth0.com/samlp/T95l6gp1PoerbniGDm9eMUaUcKiBJVZt
 - application:
     authorized_groups:
+    - team_moco
+    authorized_users: []
+    client_id: WQnlYTEc4PYRRK3FT7n3l9VOtew1IDo2
+    display: false
+    logo: auth0.png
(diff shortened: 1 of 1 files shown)


duallain (approved):
This url also errors for me - "Invalid samlResponse or relayState from identity provider"  but looks good otherwise.

No icon added to the dashboard, I assume Kevel isn't widely used so a generic icon is ok?

floatingatoll (approved):
This seems theoretically fine, assuming it works in staging.

dividehex:
> 
> 
> This url also errors for me - "Invalid samlResponse or relayState from identity provider" but looks good otherwise.
> 
> No icon added to the dashboard, I assume Kevel isn't widely used so a generic icon is ok?

Yes, the URL isn't going to work until Kevel sets it up with the provided metadata on their side.  And yes, the previous application did not have an Icon, so I'm continuing the unless otherwise directed differently.

PR #479, opened 2021-06-14
Add Everfi application [bug 1712376]

dividehex:
I'm opting to leave the display set false until this app is ready to go live.


sciurus (approved):
This looks correct to me. I verified the client id.

PR #481, opened 2021-06-21
Removing legacy Jira and putting Jira Cloud in place

rtucker-mozilla:
Removing legacy Jira and putting Jira Cloud in place


sciurus (changes requested):
@rtucker-mozilla I think you should also change `display: false` to `display: true`

sciurus (approved):

PR #483, opened 2021-06-24
(Do not merge until culture amp finish setup) Add mofo culture amp application [bug 1717822]

dividehex:
This PR adds the Mofo specific culture amp account.  Note: culture amp has not yet configured their side so URL probably won't work until it is.

### DIFF
--- a/apps.yml
+++ b/apps.yml
@@ -3845,7 +3845,6 @@ apps:
 - application:
     authorized_groups:
     - team_moco
-    - team_mofo
     - team_mozillaonline
     authorized_users: []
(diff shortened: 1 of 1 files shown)


sciurus:
After this change MoFO will have access to two different cultureamp logins? Is that expected, or should they lose access to the moco one?

dividehex:
> 
> 
> After this change MoFO will have access to two different cultureamp logins? Is that expected, or should they lose access to the moco one?
It was confirmed in the bug that mofo should lose access to the non-mofo culture amp.  The PR now reflects this.

sciurus (approved):

PR #485, opened 2021-07-06
Add BitSight to apps.yml

gene1wood:
Here's the bugzilla ticket related to this PR : https://bugzilla.mozilla.org/show_bug.cgi?id=1718342


floatingatoll (commented):
Is the URL supposed to be /samlp/metadata/ or just /samlp/?

gene1wood:
@floatingatoll really good catch, yup that was a mis copy/paste on my part. I'll commit a fix. Thank you for catching that.

floatingatoll (approved):
This seems fine now. I'm uncomfortable approving this without context (there's no description in the PR), as I can't consider whether authorized_groups or display:true are correct versus the request, so you're on your own for that.

gene1wood:
@floatingatoll ya good point, I was expecting @dividehex to take a look at it (and who has context) and I forgot to link to the associated bugzilla ticket. I've updated the PR description with the link to the bugzilla ticket.

dividehex (approved):
LGTM 👍

PR #493, opened 2021-07-28
Adding client_id for compiler explorer.

rtucker-mozilla:
Fix compiler explorer client


sciurus (approved):
I've verified this is the right client id.

PR #495, opened 2021-07-29
Add lando.devsvcdev.mozaws.net dashboard app - Bug [1720522]

dividehex:
This PR adds a dashboard application for an existing auth0 app.  It was discovered that the lando staging app had a dashboard app but the dev did not.  This fixes that.

See bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1720522

### DIFF
--- a/apps.yml
+++ b/apps.yml
@@ -2504,6 +2504,18 @@ apps:
     - team_moco
     - team_mofo
     authorized_users: []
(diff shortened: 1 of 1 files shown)


floatingatoll (approved):
Access limited to staff (groups+client_id), tile not shown. The hostname looks strange (it says dev twice) but that's probably normal. Is 2512 really the right application URL?

PR #498, opened 2021-08-03
Add new splunk dev environment, and add Netops to new & existing Splunk endpoints

pwnbus:


apps.yml
```diff
@@ -4028,6 +4028,7 @@ apps:
     authorized_groups:
     - team_secops
     - team_opsec
+    - team_netops
```
floatingatoll, line 4031: I presume secops is okay with this extension of access to a formerly-security-only group?

PR #499, opened 2021-08-04
docs: add tiles show/hide explanation to apps.yml

floatingatoll:
Hi, SSO Dashboard apps.yml PR reviewers.

I'm looking to confirm that this documentation change is clear and understandable, and that it makes sense to everyone. It's been through a couple rounds of internal review before PR, and if there are no blocker objections uncovered in this final set of review requests, I'm going to ship it to close IAM-886.

I've requested review from all current apps.yml reviewers, both primary (websre west) and last-resort/standby/concerned-bystanders. If you take the time to review this documentation-only update, I would absolutely appreciate it, and if your time doesn't permit, I understand and it's fine.

### DIFF
--- a/apps.yml
+++ b/apps.yml
@@ -1,3 +1,27 @@
+# NOTE: authorized_groups and authorized_users control two things independently:
+#
+# 1. Does the SSO Dashboard display a tile to the user for this application?
+#
+#        No, if display: false.
+#            No SSO users can see the tile.
+#
+#        Maybe, if display: true.
+#            If groups/users are defined,
+#                The user must be a member of them to see the tile.
+#            Otherwise,
+#                All SSO users (including the general public) can see the tile.
+#
+# 2. Does Auth0 restrict SSO access to this application by group or user?
+#
+#        No, if client_id is missing.
+#            All SSO users (including the general public) can access the app.
+#
+#        Maybe, if client_id is defined.
+#            If groups/users are defined,
+#                The user must be a member of them to access the application.
+#            Otherwise,
+#                All SSO users (including the general public) can access the app.
+
 apps:
 - application:
     authorized_groups:


bkochendorfer (commented):
Content looks good to me. Do you think it be worthwhile to add this to the `README`?

floatingatoll:
> Content looks good to me. Do you think it be worthwhile to add this to the `README`?

Unknown; I've never really noticed the repo's README, but I sure notice apps.yml a lot, so that's why I'm focusing on the one place we're all most likely to see it.

duallain (approved):

gcoxmoz (approved):
I can't speak to the truthfulness of the flowchart, but assuming it's correct, the flowchart reads unambiguous and complete and understandable to me.  "Certainly better than nothing, and it's added in a good location."

apps.yml
```diff
@@ -1,3 +1,27 @@
...
+#                All SSO users (including the general public) can see the tile.
+#
+# 2. Does Auth0 restrict SSO access to this application by group or user?
+#
+#        No, if client_id is missing.
+#            All SSO users (including the general public) can access the app.
+#
+#        Maybe, if client_id is defined.
+#            If groups/users are defined,
+#                The user must be a member of them to access the application.
```
gene1wood, line 21: The user must either be in the `authorized_users` list or be a member of at least one of the groups in the `authorized_groups` list. Either of these conditions satisfy the requirement.

apps.yml
```diff
@@ -1,3 +1,27 @@
...
+#            Otherwise,
+#                All SSO users (including the general public) can access the app.
```
gene1wood, line 23: All SSO users (including the general public) have [passed the first stage of access control](https://github.com/mozilla-iam/mozilla-iam/#2-stage-access-validation) and the only thing that governs user access is the second stage access control performed by the relying party.

gene1wood:
@floatingatoll oh whoops, I was late on this. I added some clarification comments above that were ambiguous

apps.yml
floatingatoll: Yeah, my wording is imperfect, but "the user is in one of the groups or the user is in the users list" starts to get weird, and it balloons from there. One of my criteria was 'fits into an 80x25 window' so that everyone reads it when they open the file in an editors, at least if they start at the top, which many do. If you feel there is an alternate interpretation that is materially problematic I'm happy to take more PRs here.

apps.yml
floatingatoll: "can access the app" isn't the best abbreviation for that. I'm not sure what is — perhaps "can continue to the app", so that we're not implying statements about app access?

PR #501, opened 2021-08-05
Add secops and opsec (security assurance) to old splunk

pwnbus:


floatingatoll (approved):
Does what it says on the tin.

PR #507, opened 2021-08-18
Change access for Contentful to a people.m.o group

pmac:
@sciurus This will require the `contentful-access` group to be created first.

### DIFF
--- a/apps.yml
+++ b/apps.yml
@@ -219,7 +219,7 @@ apps:
     - /cloudhealth
 - application:
     authorized_groups:
-    - team_moco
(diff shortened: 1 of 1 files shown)


floatingatoll:
The curators of that group, when created, will be @pmac @slightlyoffbeat, per thread on #it-web-sre.

floatingatoll:
Group is created, passes @dividehex's inspection.

floatingatoll (approved):
Matches the group and looks correct. Has a client_id, so the restriction will apply.

PR #508, opened 2021-08-19
Update SiQ (previously SpaceIQ) - Bug [1725161]

dividehex:
This PR removes the preview tile, updates the logo on the "true" tile and enables displaying it with the new name "SiQ"


floatingatoll (approved):
Looks good, and resolves the "two SpaceIQ" question I sent out a few days ago. Thank you for doing this!

PR #510, opened 2021-09-07
Add coderpad.io and remove hackerrank - Bug [1726256]

dividehex:
Coderpad.io replaces hackerrank.  The 2 authorized users are contractors and therefore need to be explicitly authorized here.


apps.yml
```diff
@@ -3369,15 +3369,17 @@ apps:
     - hris_is_staff
     - team_moco
     - team_mofo
-    authorized_users: []
-    client_id: 1B1ser8z1eyr4nOFtW1acz01i8GYIaBo
+    authorized_users:
```
floatingatoll, line 3372: I suppose nowadays I would consider requesting an override group that we can hand over ownership/maintenance of back to the requesting party, but this is still fine.

floatingatoll (approved):
Could you fix the bug number in the PR title before merging?

PR #513, opened 2021-09-09
Add admin role accounts to Box [#1729527]

floatingatoll:
As requested in bug.


duallain (approved):
Looks good, valid way to define lists of users. User are specified with full email as authorized_users for a few other blocks in this file. Makes some sense that a few service accounts would need access for box.

PR #514, opened 2021-09-16
Add team_elance to udemy/getabstract [SE-2617]

floatingatoll:


gcoxmoz (approved):
`team_elance` added where `team_moco_benefited` existed already.  Seems like the right idea for this request.

PR #516, opened 2021-09-30
Apps: add settings.dev.mozaws.net(kinto-dev) app

rvaralda:
add settings.dev.mozaws.net(kinto-dev) app

### DIFF
--- a/apps.yml
+++ b/apps.yml
@@ -4144,3 +4144,15 @@ apps:
     url: https://auth.mozilla.auth0.com/samlp/SgKK7ZZcUT8gLcZPYq229gCHBIyWV2kq
     vanity_url:
     - /bitrise
+- application:
+    authorized_groups:
+    - hris_is_staff
+    - team_moco
+    - team_mofo
+    authorized_users: []
+    client_id: 7M49iYODCiCGdldu4awmF2Pr6pHsRZDe
+    display: false
+    logo: auth0.png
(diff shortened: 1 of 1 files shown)


apps.yml
```diff
@@ -4144,3 +4144,15 @@ apps:
...
+    - hris_is_staff
+    - team_moco
+    - team_mofo
+    authorized_users: []
+    client_id: 7M49iYODCiCGdldu4awmF2Pr6pHsRZDe
+    display: false
+    logo: auth0.png
+    name: settings.dev.mozaws.net
+    op: auth0
+    url: https://settings.dev.mozaws.net/v1/admin/
```
floatingatoll, line 4158: Should this be the auth0/samlp URL with the client_id in it?

rvaralda: In this case I just keep the same config as other environments.
i.e. https://github.com/mozilla-iam/sso-dashboard-configuration/blob/master/apps.yml#L2525

floatingatoll:
Ah, since there’s no redirect shorturl, and no tile displayed, the URL
probably isn’t used at all. It should be fine then.

floatingatoll (approved):

bkochendorfer:
@rvaralda commit looks good but you need to sign it https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits

PR #525, opened 2021-10-28
Add fivetran to dashboard [#1737182]

dividehex:
Simple dashboard add


floatingatoll (approved):
If moco-only is acceptable, then this is acceptable. I'm not sure if hris_is_staff is a current affair, but cargo-culting it does no harm.

PR #526, opened 2021-11-02
Internal Pocket Admin Tools [#1730268]

dividehex:
Confirmed via slack, `moco` is the only group that should have access.


dividehex:
> I don't like the idea of _all_ of MoCo having admin access to Pocket, but it's not my place to obstruct that decision here. I would strongly encourage consulting someone with Security or Risk in their title before going live with this, if you have not already done so.

That is a good point.  I was pointed at a previous tool that had similar scope so I assumed the wide scope was on purpose and double checked on slack but I'll dig in further before proceeding.

dividehex:
Confirmed there are permissions manually handled on the application side so I am going to proceed to merge this.

PR #533, opened 2021-12-09
Add sentry.io [#1742454]

dividehex:


duallain (approved):
LGTM

just to be doulbe paranoid I checked the whole file for the clientid listed here and there are no duplicates. And 'sentry' is unique. So code should be good on its own, and not collide with the existing config.

rvaralda (approved):

PR #534, opened 2021-12-10
Limit stripe_subplat access and display tile

dividehex:
This PR narrows access to the Stripe subplat account and displays the Stripe tile on the dashboard


dividehex:
@gcoxmoz Would you mind verifying the LDAP groups listed in this PR?  I noticed a typo in the bug comment I copied them from so I just want to be sure the typo wasn't copied over to LDAP also.

apps.yml
```diff
@@ -3712,12 +3712,14 @@ apps:
     url: https://stats.voice.mozit.cloud/
 - application:
     authorized_groups:
-    - team_moco
-    - team_mofo
+    - stripe_subplat_admin
+    - stripe_subplat_analyst
+    - stripe_subplat_developer
+    - stripe_subplat_supportsp
```
gcoxmoz, line 3718: stripe_subplat_viewonly was added in https://bugzilla.mozilla.org/show_bug.cgi?id=1637117#c33 so I think you want that here also?

gcoxmoz (changes requested):

apps.yml
dividehex: @gcoxmoz Thanks for catching that! 💯

gcoxmoz (approved):
Group list makes sense, display change makes sense.  Based on the comments at the top of the file I think this does what you expect.  I have no reason to think this is a bad change.

PR #535, opened 2021-12-13
Add team_mozillaonline to sentry.io

dividehex:


duallain (commented):
I'm not sure how to review the underlying goal of this pr - should all users in mozilla online get access to sentry?  Is there a ticket where we've communicated with the services sres to see if this is something they want?  

The actual change lgtm, assuming this is what we want.

dividehex:
@duallain All apologies for lack of a bug number. I did this in haste.  I have confirmed with the requestor (:wezhou) via slack that this is required.
See https://bugzilla.mozilla.org/show_bug.cgi?id=1742454

duallain (approved):
++ Wei says it's good, it should be good.

PR #536, opened 2021-12-16
Add new callback URL for Kinto DEMO

rvaralda:
[SE-2947](https://mozilla-hub.atlassian.net/browse/SE-2947)
https://bugzilla.mozilla.org/show_bug.cgi?id=1743115

### DIFF
--- a/apps.yml
+++ b/apps.yml
@@ -4117,4 +4117,4 @@ apps:
     logo: auth0.png
     name: demo.kinto-storage.org
     op: auth0
-    url: https://settings.dev.mozaws.net/v1/admin/
+    url: https://demo.kinto-storage.org/v1/admin/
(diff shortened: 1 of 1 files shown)


smarnach:
This used to work for the old demo server, if I remember correctly, so I think there must be some existing configuration that we can change, rather than adding a completely new one.

dividehex (approved):
Since technically this is a new instance of the Kinto service, it would probably be a better practice to create an new RP with a new Client ID.  But since the existing service was never listed in the apps.yaml and the previous instance looks to be gone, we can just update the existing RP.

I've checked the for duplicate clientIDs and name.  All looks good.

smarnach:
@dividehex Where does the existing configuration live? Was it manually created in the web interface?

PR #542, opened 2022-02-14
Drop old Splunk from apps.yml [NETOPS-2781]

ctbfourone:


dividehex (commented):
Before removing this metadata, the Auth0 app will need to be deleted first.  Then we can land this.

PR #545, opened 2022-02-17
A simple ruby tool to check if we have multiple configurations for th…

duallain:
…e same clientid

# All PRs are assigned reviewers by default; selecting your own is optional.


dividehex (approved):
👍 nice add

PR #548, opened 2022-02-22
Add valimail [IAM-920]

dividehex:
This PR adds an app entry for Valimail


floatingatoll (approved):

dividehex:
I jumped the gun on this one. It looks like Valimail already has SSO enabled and exists in the app. 🥇

PR #549, opened 2022-03-09
Add modernhealth [1757773]

dividehex:
~Before merging this, I need to confirm `authorized_groups` is set to the correct scope.~

Group scope is correct.


dividehex:
Also needs an image

apps.yml
```diff
@@ -4018,3 +4018,15 @@ apps:
     url: https://auth.mozilla.auth0.com/samlp/FoeZGnqq0X9fQXQQ9Ve5kdSdPN4eedlG
     vanity_url:
     - /prefect-pocket
+- application:
+    authorized_groups:
+    - team_moco_benefited
```
floatingatoll, line 4023: Does mofo expect to have access to this as 'benefited employees' in general? If so, team_mofo, I believe, is all that's required (they don't differentiate benefited beyond that).

floatingatoll (approved):
r+ with caveat noted about mofo

floatingatoll:
Consider fixing the typo in [de0c2fb](https://github.com/mozilla-iam/sso-dashboard-configuration/pull/549/commits/de0c2fba13d0cdf21ea3ef011e9d445b2c6fa558) description as well, though not strictly necessary.

apps.yml
dividehex: I'll double check on whether mofo should be included

dividehex: MoFo will not be covered.  It’s a completely different legal entity and they offer their own set of benefits.

floatingatoll:
Replaced by #559.

PR #550, opened 2022-03-09
Add logikcull [1757780]

dividehex:
This PR add an SSO app entry for Logikcull.


floatingatoll (approved):
I would be curious what the 404 rate for /logickull is over time, but this is structurally sound, regardless :)

PR #564, opened 2022-04-20
Bug 1637172 - Remove all `hris_is_staff` rules from apps.yml

l-hedgehog:
Per action item 1 in [comment #9](https://bugzilla.mozilla.org/1637172#c9)


gcoxmoz (approved):
Does what it says on the tin: all the `hris_is_staff` are gone, no `authorized_groups` get emptied out so no fundamental changes.  Seems right to me.

floatingatoll (approved):
I concur: this is a flat removal of the gone-away LDAP group and no issues should occur from its removal.

PR #597, opened 2022-05-26
Copy all team_mofo permissions to team_mzla

floatingatoll:
Grant team_mzla the same permissions as team_mofo, as part of onboarding MZLA-of-MoFo with the same access as MoFo.

### DIFF
--- a/apps.yml
+++ b/apps.yml
@@ -53,6 +53,7 @@ apps:
     authorized_groups:
     - team_moco
     - team_mofo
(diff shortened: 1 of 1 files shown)


gcoxmoz (approved):
Downloaded the proposed file, `cat /tmp/proposed | grep -A1 team_mofo | sort | uniq -c` says 191 `team_mofo`'s and 191 `team_mzla` lines.  Looks right.

PR #598, opened 2022-06-02
Add office support staff to apps they need. IO-2052

gcoxmoz:
The office support folks (currently receptionists + mailroom) are currently
manually misclassified as team_moco to get them access to essential tools.
In order to remove such forced errors and their knock-on effects, I've created
a group for them to belong to, so that they may be only get what access
their roles require.

### DIFF
--- a/apps.yml
+++ b/apps.yml
@@ -381,6 +381,7 @@ apps:
     - team_mozillajapan
     - team_mozillaonline
     - travelling_accounts
+    - team_office_support
     - gsuite_shared_accounts
     - moc_service_accounts
     authorized_users:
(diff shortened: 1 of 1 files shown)


floatingatoll (approved):
This is a structurally sound commit, but I'm not qualified to say whether this is the complete list of services they'll need. Approving so that cleanup work can move forward, and I assume we'll have minor patch fixes over time if anything is missed.

gcoxmoz:
> I assume we'll have minor patch fixes over time if anything is missed.

Exactly the intention.  We covered the big apps that are known, and we expect them to gripe to servicedesk if we missed something.

dividehex:
@gcoxmoz Can you resubmit this with your commits signed?  I'm not able to merge unsigned commits.

PR #600, opened 2022-06-10
add /people shortcut to people.m.o

ctbfourone:


floatingatoll (approved):
Make sure it works once it's live, but I think it probably will at least be mostly harmless.

PR #603, opened 2022-07-07
Hide the riskheatmap on the dashboard

gene1wood:
@moz-hwine Here's that change you'd requested.


gene1wood:
Ah hadn't realized that each entry needs a logo even if it isn't displayed on the dashboard (odd). I'll close this and open a new PR

PR #608, opened 2022-07-11
Add details to README about CI process

gene1wood:
Also, wrap text so that it's possible to more easily see what's changing (instead of any change in a section showing the entire section changing)


ctbfourone (approved):
+1, documentation improvements!

PR #611, opened 2022-07-26
Update phpldapadmin configurations [SE-3246]

gcoxmoz:
This is a 2-piece PR:
Commit 1 removes a superfluous definition of ldapadmin1's phpldapadmin configuration that shouldn't have been in place.  This was discovered in the process of researching this work.

Commit 2 makes a clone of the real phpldapadmin configuration, aiming it at a new server.  Once service work is completed, the old instance will be removed.

### DIFF
--- a/apps.yml
+++ b/apps.yml
@@ -704,6 +704,16 @@ apps:
     logo: auth0.png
     name: PHPLDAPAdmin
     op: auth0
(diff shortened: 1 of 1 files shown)


ctbfourone:
+1, the first diff was weird to read, but, that's diffs sometimes. looks good.

ctbfourone (approved):
+1

PR #612, opened 2022-08-02
Cleanup of phpldapadmin entry [SE-3246]

gcoxmoz:
This is a followup to #611.
Commit 1 removes the dupe created in that PR;  Commit 2 changes our URL to the host instead of the long ugly SAML URL, as I have pivoted from SAML to OIDC on ldapmanager1.

PR #613, opened 2022-08-25
Cleanup of phpldapadmin entry [SE-3246]

gcoxmoz:
Dupe of #612, but with signed commits, sorry.

This is a followup to #611 .
Commit 1 removes the dupe created in that PR; Commit 2 changes our URL to the host instead of the long ugly SAML URL, as I have pivoted from SAML to OIDC on ldapmanager1.


ctbfourone (approved):
+1, looks legit, I get redirected to auth as expected.

PR #614, opened 2022-08-26
Add travelling_accounts group to support users made for high-threat travel

gcoxmoz:
This group addition comes from work to support https://mana.mozilla.org/wiki/display/SECURITY/Security+Precautions+when+traveling - wherein we want to give users going into high-infosec-threat countries a temporary user who can do minimal tasks.

In this case, the list was settled at to be Zoom, Slack, gmail/gcal/gsuite.  This list was reviewed by EUS to be the minimum needs to start from.

### DIFF
--- a/apps.yml
+++ b/apps.yml
@@ -380,6 +380,7 @@ apps:
 #    - team_mzla
     - team_mozillajapan
     - team_mozillaonline
+    - travelling_accounts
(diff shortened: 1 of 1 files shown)


gene1wood (approved):
:+1:

ctbfourone:
+1, looks good.

PR #621, opened 2022-09-20
update to pocket-slack

ctbfourone:
- new icon
- update app name and vanity_url

This is already a visible tile, but afaict it is still unused because SSO hasn't been configured on Pocket's Slack instance. So this is just some pre-work before we do enable it.


ctbfourone:
I will reopen this.

PR #627, opened 2022-10-12
add cbrentano@mozilla so I can test SSO for this app

ctbfourone:
When doing a final check for enabling SSO on Pocket's Slack instance I realized that I would be locked out since I'm not a member of the authorized_groups, so this is just a temporary bodge to provide my LDAP user access so I can test things, afterwards I will revert this.


ctbfourone:
Taking a different route.

PR #630, opened 2022-10-19
add shepherd service (dev/stage/prod) and generic logo image

ctbfourone:


dividehex:
Do all 3 tiles need to be displayed?  I would think only production would need to be on display.

ctbfourone:
Yeah true, but I'll have to redo this because there's a conflict, I tried to fix it, then rebase didn't sign my commit. Whee.

PR #631, opened 2022-10-19
DO NOT MERGE until 10/24 - Confluence [1793637]

dividehex:
This PR sets up a new tile for Confluence although hidden for now. One thing to note here is Confluence and Jira share the same SSO integration since they are both part of the same Atlassian account.  Access to the various apps will need to be managed within Atlassian on a per user basis.

https://bugzilla.mozilla.org/show_bug.cgi?id=1793637


ctbfourone (approved):
+1, lgtm!

PR #628, opened 2022-10-17
Allow zoomadmin to access jira, bug 1795163

gcoxmoz:
https://bugzilla.mozilla.org/show_bug.cgi?id=1795163

PR #634, opened 2022-10-25
Add zoomadmin to JSM, 1795163

gcoxmoz:
This is a correction from #628 based on new information.


ctbfourone (approved):
+1, LGTM

gcoxmoz:
Forced signed commits, without a hook to block us from mistakes, sucks.

PR #635, opened 2022-10-25
Add zoomadmin to JSM, 1795163

gcoxmoz:
This is a correction from #628 based on new information.
#634 was this, but not signed.


ctbfourone (approved):
+1, with feeling

PR #637, opened 2022-11-01
Add valimail [IAM-920]

floatingatoll:
This is a re-submit of #548 with the client_id corrected.


floatingatoll:
(We should hold on merging this until the correct set of users are in the mozilliansorg group.)

dividehex:
There is already an entry for client id `wXpnkKIgoucVrybZmxiPnnpeSqn816qD` at line 3750.  That should probably be remove in favor of adding this.

floatingatoll:
Agreed, good catch.

floatingatoll:
Updated to reflect that.

dividehex (approved):
r+ lgtm

PR #642, opened 2022-12-13
Add team_gmail_only to access gmail

gcoxmoz:
Per an odd request, we'd like to have a group of users who
* aren't team_moco / other autogroup members,
* can access gmail
* can't access gcal/gdocs.

... and here we are.


dividehex (approved):
r+

PR #644, opened 2022-12-20
Add Compass

ctbfourone:


gcoxmoz:
Linting nitpick: your YAML indents `authorized_groups` and `vanity_url` where other entries don't.

And, "display is false".  Nothing wrong with that, just mentioning since you have a tile graphic attached, and I don't see an indication of planning for showing the tile in the future.

ctbfourone:
Ah thanks, fixed the indentation. `display: false` can be toggled later if needed, I mostly wanted to get this in there so that we can scope access to just `team_moco` and `team_mofo`. 

Also yeah, the icon is Not Great, but it was the best thing I could scrape from the site at a glance. :)

bkochendorfer (approved):

PR #645, opened 2023-01-19
imageoptim all SSO tiles, saving 36% of 2.2MB [IAM-933]

floatingatoll:
No changes to image resolution or EXIF tags, just an imageoptim pass to apply lossless reductions. All images should be pixel-perfect unchanged, and still open correctly in browsers based on a few random spot tests.


gcoxmoz (approved):
File preview in GH blows out trying to diff up this many images, so, this approval is somewhat a shot in the dark.  But given the audible from atoll on what she did to generate these, and that the early ones all look cromulent, I see no reason to worry about the unseen ones.

ctbfourone (approved):
+1, all looks good to me

PR #646, opened 2023-01-23
remove legacy PTO apps [IAM-1057]

floatingatoll:
Please confirm in IAM-1057 that it's OK to proceed with the owners, and that all listed Auth0 applications have been removed, before committing and releasing.

(HOLD until 1/31)

### DIFF
--- a/apps.yml
+++ b/apps.yml
@@ -1141,28 +1141,6 @@ apps:
 - application:
     authorized_groups:
     - team_moco
-    authorized_users: []
-    client_id: Q3z1fjeoZhGyws1IXDUc6rHdcYNpxTv8
-    display: false
-    logo: auth0.png
-    name: PTO
-    op: auth0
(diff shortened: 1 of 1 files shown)


ctbfourone (approved):
+1

gcoxmoz (approved):
The decom of PTO is slightly premature in one sense (we're going to nuke it in the near future), and late in another (the existing linkage is broken because old PTO is disabled).  But pokes at the migration PMs for "if you want SSO to move, you need to provide a link / integration" have not been responded to.  If they wish to have a new endpoint, then we should take it up as new work, and not hold up this cleanup.  r+

floatingatoll:
(Note that we can't _remove_ the entry until it's torn down in Auth0, which has to wait for the 1/31 full decom in a ~week, so the hold remains until OK'd to proceed in ticket.)

PR #647, opened 2023-02-01
Add salesforce cloud logo, IAM-1061

gcoxmoz:
Salesforce logo determined to be missing, adding.
I'm terrible at graphics, so if this is bad someone deny me, but I don't know anything better.


ctbfourone (approved):
☁️

PR #656, opened 2023-02-17
Add NetSuite

ctbfourone:


apps.yml
```diff
@@ -4182,3 +4182,15 @@ apps:
     url: https://auth.mozilla.auth0.com/samlp/TPliDroeFM4p3csAsmC0tfF8AxeOisrP
     vanity_url:
     - /compass
+- application:
+    authorized_groups:
+      - mozilliansorg_netsuite-access
```
gcoxmoz, line 4187: If you're feeling linty, dedent this a couple of spaces

gcoxmoz (approved):
Pattern matches enough other entries to make sense.
https://bugzilla.mozilla.org/show_bug.cgi?id=1804908 for archaeologists.

apps.yml
ctbfourone: Oh good catch, stupid text editor.

PR #505, opened 2021-08-13
Group certain complicated SSO entries together [SE-2440]

floatingatoll:
While working on SE-2440 we noticed that complicated SSO entries are spread out throughout the file. This PR attempts to improve that somewhat and includes comments from our analysis and intentions.

This should have zero functional changes to any service listed, and I did my best to retain the ordering in case that's secretly relevant.

PR #657, opened 2023-02-17
Redo: Group certain complicated SSO entries together [SE-2440]

ctbfourone:
A redo of https://github.com/mozilla-iam/sso-dashboard-configuration/pull/505 due to age/conflicts/etc.

> While working on SE-2440 we noticed that complicated SSO entries are spread out throughout the file. This PR attempts to improve that somewhat and includes comments from our analysis and intentions.
> 
> This should have zero functional changes to any service listed, and I did my best to retain the ordering in case that's secretly relevant.

### DIFF
--- a/apps.yml
+++ b/apps.yml
@@ -1208,22 +1208,6 @@ apps:
     url: https://admin.readitlater.com
 - application:
     authorized_groups:
-    - team_moco
-    - team_mofo
-#    - team_mzla
-    - team_mozillaonline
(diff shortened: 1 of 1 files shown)


gcoxmoz (approved):
Each commit appears, to my read, to be a bug-for-bug-identical block-shift of entries to put similar RPs together for human readability / finding.   "Does what it says on the tin."

PR #670, opened 2023-04-05
Boxdotcom

ctbfourone:


gcoxmoz:
https://bugzilla.mozilla.org/show_bug.cgi?id=1822333

This is adding a phonebook group on top of moco/mofo/nda.. which should be 'nearly everyone' to begin with.  That's slightly odd, so I wanted to see if the plans were considered.

It'll take 'a while' (30d, an auth0 refresh cycle) to have all users show up as being `boxcom-access` members, so the overlap helps there.  BUT, when you land this, the tile will be visible to all those people, so folks might start clicking in and gumming up the works.  In my (backseat-driving) opinion I'm thinking, break this up into:

* fix the picture, add the authorized group
* wait 30d
* make the tile visible to all and drop the other groups.

But I could be missing things.

ctbfourone:
As I understand it, only people with Box accounts are able to login anyway. Originally an LDAP group was requested but I steered that towards a phonebook group instead. I'll ask James to weigh in here, but my take is that the overlap will give SD and AE time to get people added to the group, and once that's done we can later pull `team_*` from the authorized groups list.

data-sync-user:
James here. This sounds good--takes care of adding all the existing users and displaying the Box tile after the ~30 day estimated refresh period.

floatingatoll (approved):
This seems like a plausible, normal PR and has a valid set of changes.

ctbfourone:
Thank you!

PR #671, opened 2023-04-12
Update Remote Settings DEV URL

leplatrem:


grahamalama:
@gcoxmoz https://mozilla-hub.atlassian.net/browse/OPST-962

gcoxmoz (approved):
Tweaking an existing integration, ownership seems right, we have prior art of using allizom in this config.  LGTM.

apps.yml
```diff
@@ -3638,7 +3638,7 @@ apps:
     logo: auth0.png
     name: settings.dev.mozaws.net
```
gcoxmoz, line 3639: ugh.  I click approve right as I see something.
Should the name line change to match the URL?

leplatrem:
@gcoxmoz this should be ok now

PR #672, opened 2023-04-14
Add hireEZ [IAM-1111]

dividehex:
PR adds new sso entry for HireEZ


gcoxmoz (approved):
Pattern matches other apps, looks right.

PR #673, opened 2023-04-25
Add entry for Auth0 user management app [IAM-1122]

dividehex:
This PR adds entries for the dev and prod instances of the Auth0 user management app


gcoxmoz (approved):
Looks reasonable for a new spinup

PR #679, opened 2023-05-17
Add Pocket's dbt Cloud

ctbfourone:
The groups stanza may be redundant, but there is a distinction between the two groups: team_moco will have read-only access, while pocket-dbt-developer-access will have developer access. Putting both in leaves some flexibility if one or the other needs changing.

https://mozilla-hub.atlassian.net/browse/IAM-1135


apps.yml
```diff
@@ -4248,3 +4248,16 @@ apps:
     url: https://auth.mozilla.auth0.com/samlp/VezzSFjr5YxbFicOqhLmnL8UfUGALQtl
     vanity_url:
     - /notion
+- application:
+    authorized_groups:
+    - team_moco
```
gcoxmoz, line 4253: nonblocking review: should this be `team_pocket`?

ctbfourone: Ah good catch, I'll add.

gcoxmoz: `team_pocket` is a manually curated subset of `team_moco`.  It's "folks who are pocket" and we have never been able to determine who those folks are from workday info.
If you have `team_moco` you don't need `team_pocket`.  If you need more than `team_pocket` then stay with `team_moco`.  This comment is an "I don't know who all needs it" but it's just to call attention to `team_pocket` existing.  "Choose one, your call."  Sorry for not being clear to start from.

dividehex (approved):

PR #680, opened 2023-05-19
Add Mode analytics for Pocket

ctbfourone:
https://mozilla-hub.atlassian.net/browse/IAM-1139


gcoxmoz (approved):
I'm approving based on 'looks reasonable' / 'matches existing patterns' rather than any deep knowledge on the quality, but "new service" so there's no impact if there's badness.

PR #683, opened 2023-05-26
removed moderator and updated vanity url

vansterzzz:
IAM-1130. removed old moderator app and updated vanity url to duplicate moderator app

### DIFF
--- a/apps.yml
+++ b/apps.yml
@@ -603,18 +603,6 @@ apps:
     vanity_url:
     - /mdn
 - application:
-    authorized_groups:
-    - everyone
-    authorized_users: []
-    client_id: nPr8QRo0dLxM3RRHwIXRSvhmOSPDvNr4
-    display: true
-    logo: moderator.png
-    name: Moderator
-    op: auth0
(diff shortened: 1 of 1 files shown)


apps.yml
```diff
@@ -602,18 +602,6 @@ apps:
     url: https://developer.mozilla.org/
     vanity_url:
     - /mdn
-- application:
-    authorized_groups:
-    - everyone
-    authorized_users: []
-    client_id: nPr8QRo0dLxM3RRHwIXRSvhmOSPDvNr4
-    display: true
```
gcoxmoz, line 610: display:true and logo:moderator.png are going away; the surviving entry is display:false and logo:auth0.png.  Should these transfer over too?

gcoxmoz (approved):
I'm approving these based on the discussion we had in triage; this eliminates an unused clientid and preserves a valid clientid and puts the vanityurl over onto it.  The concept is sound enough to approve.  The tweaks of "do we want to make the tile visible" are somewhat outside my realm of knowledge, but I don't think it's worth blocking over.  It's a low-pri RP, so worst case, it becomes some editing later.

PR #685, opened 2023-06-06
allowed team_mzla into udemy

vansterzzz:
IAM-1147, enabled team_mzla access into Udemy app.


gcoxmoz (approved):
Change does what it says on the tin, and IAM-1147 covers that it's been through a legal review since it's offering benefits to a subsidiary.

PR #686, opened 2023-06-08
Swap khoros client IDs

dividehex:
This PR swaps the IDs used for Prod and Stage.  It matches reality due to some changes made on the SP side.  This  change is mostly to makes sure the vanity urls match properly to their respective environments.

### DIFF
--- a/apps.yml
+++ b/apps.yml
@@ -3672,12 +3672,12 @@ apps:
     authorized_groups:
     - everyone
     authorized_users: []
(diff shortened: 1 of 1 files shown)


gcoxmoz (approved):
Did not verify the original statement of these being crossed up but it's a believable assertion, the ID swaps match each other, so no loss of information here.  Approving based on reasonable trust.

PR #687, opened 2023-06-09
removed unnecessary groups from boxcom-access

vansterzzz:
iam-1148, removed team_mo[c,f]o and mozillians_nda

### DIFF
--- a/apps.yml
+++ b/apps.yml
@@ -1017,10 +1017,7 @@ apps:
     - /iam-grafana
 - application:
     authorized_groups:
-    - mozilliansorg_nda
     - mozilliansorg_boxcom-access
-    - team_moco
-    - team_mofo
     - team_office_support
 #    - team_mzla
     authorized_users:


apps.yml
```diff
@@ -1017,10 +1017,7 @@ apps:
     - /iam-grafana
 - application:
     authorized_groups:
-    - mozilliansorg_nda
     - mozilliansorg_boxcom-access
-    - team_moco
-    - team_mofo
     - team_office_support
 #    - team_mzla
```
gcoxmoz, line 1022: Nitpick, let's strike the mzla commented line.  The purpose of this PR is to remove team_moco type groups.  The commented line indicates mzla might ever get added... which is now wrong.

gcoxmoz (changes requested):

dividehex (approved):

apps.yml
dividehex, line 1022: I agree. `team_mzla` should be removed from here completely since access it covered explicitly by `mozilliansorg_boxcom-access`

dividehex (changes requested):

dividehex (approved):
LGTM

PR #690, opened 2023-06-22
Update sso.allizom.org client ID

dividehex:
This updates the sso.allizom.org client ID to properly match the client application in the Auth0 dev tenant.


gcoxmoz (approved):
allizom = nonprod = blindly approving.

PR #692, opened 2023-06-27
Update BitSight access control

gene1wood:
This changes the access control for BitSight so that it matches that access control used within the BitSight vendor. This way a user only need be added to, for example, the `bitsight-users` people.mozilla.org access group to get access to BitSight (instead of how it works currently where the user needs to be a member of both `bithsight-users` in people.mozilla.org and `team_opsec` in LDAP)

### DIFF
--- a/apps.yml
+++ b/apps.yml
@@ -3368,8 +3368,8 @@ apps:
     - /everfi
 - application:
       authorized_groups:
-          - team_opsec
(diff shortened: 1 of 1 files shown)


gcoxmoz (approved):
More specific permissions than the overloaded old-name team groups?  yes plz.

PR #695, opened 2023-07-13
Add pre-commit and fix all trailing ws and end-of-file

dividehex:
This PR adds a pre-commit config and fixes all of the trailing white spaces and missing EOF chars.


gcoxmoz (approved):
Lintage looks good.  If this breaks anything, we deserve to lose.

PR #698, opened 2023-07-20
updated URL for Recorded Future

vansterzzz:
IAM-1174
updated URL for Recorded Future


gcoxmoz (approved):
I don't see the request within the listed IAM ticket, but IAM-1168 lists enough breadcrumbs to where this seems like an "ok, I see where you're going."

Approving based on 'seems legit upon analysis' rather than 'actual detailed request that's listed.'

PR #699, opened 2023-07-24
removed app paradigm reach

vansterzzz:
IAM-1176, removed app paradigm reach from apps.yml

PR #700, opened 2023-07-24
removed paradigm reach tile

vansterzzz:
IAM-1176 removed paradigm reach tile.


gcoxmoz (approved):
Hand-in-hand with #699.

PR #701, opened 2023-07-28
Remove duplicate entry for New Relic SRE

dividehex:
Removes the dup entry for New Relic IT SRE


gcoxmoz (approved):
from Slack: "Since the Moco and Mofo entry follow the SRE only, they were never being evaluated."  That obviates my concern about this casual-read-appearing to go from all-company to just-SRE.

PR #702, opened 2023-07-31
decommissioned app ideascale

vansterzzz:
IAM-1170 deleted tile and app idealscale


gcoxmoz (approved):
Looks right, removes image, lines up with Jira issue.

PR #704, opened 2023-09-07
Aws sso

dividehex:
This PR adds 3 separate tiles to the dashboard that represent the 3 unique AWS parent billing accounts.


apps.yml
```diff
@@ -4092,3 +4092,92 @@ apps:
...
+    - vpn_sumo_aws_devs
+    authorized_users: []
+    client_id: JR8HkyiM2i00ma2d1X2xfgdbEHzEYZbS
+    display: true
+    logo: aws.png
+    name: AWS (IT)
+    op: auth0
+    url: https://mozilla-aws.awsapps.com/start
+    vanity_url:
+    - /aws-it
```
floatingatoll, line 4140: We'll eventually want to use /aws here, but we don't need to _yet_.

floatingatoll (approved):
This is fine. Let your manager know when you push this since it'll advertise the new AWS access method to the people we're going to be emailing.

PR #705, opened 2023-09-11
fixed Dualo URL in apps.yml

vansterzzz:
updated Dualo URL in apps.yml


ctbfourone (approved):
+1, LGTM, I correctly auth'd and got the error that I don't have Dualo access but that was expected so, :shipit:

PR #708, opened 2023-10-10
remove two decommissioned auth0 apps

floatingatoll:
While reviewing for another task, I noticed two apps that have invalid client_ids. I don't have context for their decom bugs, but here's a PR to remove them from apps.yml for the team to review and merge.

### DIFF
--- a/apps.yml
+++ b/apps.yml
@@ -1436,17 +1436,6 @@ apps:
     authorized_groups:
     - everyone
     authorized_users: []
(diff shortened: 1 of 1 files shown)


gcoxmoz (approved):
aus4-admin has a decom citation in IO-2300.  clear r+
sso.allizom.org is allizom and by definition "if this breaks something prod, we deserve to lose" so if someone's research says it's fine to delete, I can't see standing in the way.

PR #720, opened 2023-11-27
Remove unneeded commented-out team_mzla entries

floatingatoll:
When I added these entries, we were operating under an early understanding of MZLA that was altered later on in 2022. Now that things have settled out (and it's a year later), submitting this to remove all of the unnecessary commented-out entries, which should improve the file's usability somewhat.

### DIFF
--- a/apps.yml
+++ b/apps.yml
@@ -79,7 +79,6 @@ apps:
     authorized_groups:
     - team_moco
     - team_mofo
-#    - team_mzla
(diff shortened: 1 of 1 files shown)


gcoxmoz (approved):
Concur.  These represent an onboarding placeholder... but they've been onboarded and haven't asked for changes, so yeah, zap.

dividehex (approved):

PR #722, opened 2023-12-07
Fix up moderator tile

dividehex:
This PR adds the proper tile icon to moderator and fixes the link on the dashboard to send users to the proper login screen.


ctbfourone (approved):
+1

PR #723, opened 2023-12-11
HOLD/WIP: Stage app entries for IAM-1231

ctbfourone:
For team_mzai and team_mzvc


gcoxmoz (approved):
I'm r+'ing as-is; I suspect you'll have an extra commit for confluence but rubberstamping early seems okay.

PR #726, opened 2024-01-03
Remove three retired Salesforce entries

ctbmozilla-admin:
IAM-893


gcoxmoz (approved):
Lines up with what's in the bug, bug shows diligence of outside confirmation.

PR #727, opened 2024-01-10
remote app

vansterzzz:
iam-1274, onboarding remote.com app


apps.yml
```diff
@@ -4097,3 +4097,13 @@ apps:
...
+- application:
+    authorized_groups:
+    - mozilliansorg_remote-access
+    authorized_users: []
+    cliend_id: 8kBFdSJ7MgTkqTdNQFUDzgbVK5eViSlo
+    display: true
+    logo: remote.png
+    name: Remote
+    op: auth0
+    url: https://remote-com.us.auth0.com/login/callback?connection=ba2d0fae-8ae4-488c-a2c0-85c8d5517855
```
ctbfourone, line 4109: I think usually you want the IdP-initiated sign-in URL, unless the app doesn't support that. So in that case it would be https://auth.mozilla.auth0.com/samlp/8kBFdSJ7MgTkqTdNQFUDzgbVK5eViSlo -- but I could be wrong in this case.

ctbfourone: I think in general it might be good to call this "Remote.com" because "remote" is such a generic term. So like the people.m.o. group could be `remote-com-access`, and the logo `remote-com.png`, etc.

apps.yml
```diff
@@ -4097,3 +4097,13 @@ apps:
     name: Birdbox Demo - Production
     op: auth0
     url: https://birdbox.mozilla.com/
+- application:
+    authorized_groups:
+    - mozilliansorg_remote-access
+    authorized_users: []
+    cliend_id: 8kBFdSJ7MgTkqTdNQFUDzgbVK5eViSlo
```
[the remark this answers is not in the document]
vansterzzz: ahh i copied the previous stanza (birdbox) so i fixed both cliend_id to client_id and changed remote to remote.com

apps.yml
vansterzzz: ahh i copied the previous stanza (birdbox) so i fixed both cliend_id to client_id and changed remote to remote.com

vansterzzz: >I think usually you want the IdP-initiated sign-in URL, unless the app doesn't support that. So in that case it would be https://auth.mozilla.auth0.com/samlp/8kBFdSJ7MgTkqTdNQFUDzgbVK5eViSlo -- but I could be wrong in this case.

i've had issues in the past using that format and jake told me to use what they provide as a URL. let me ask jake how i should proceed then. thanks for the review!

ctbfourone (approved):
+1, lgtm

PR #728, opened 2024-01-11
updated authorized group remote.com

vansterzzz:
updated user group for remote.com


ctbfourone (approved):
+1, looks good and matches pmo group

PR #729, opened 2024-01-12
IAM-896: Expensify apps.yml entries are not in a valid state

ctbfourone:
Removes the decommed Pocket Expensify entry and a redundant entry for the primary one.

IAM-896

### DIFF
--- a/apps.yml
+++ b/apps.yml
@@ -3714,18 +3714,6 @@ apps:
     - team_mofo
     authorized_users:
     - billing@mozilla.com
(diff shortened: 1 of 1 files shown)


gcoxmoz (approved):
Removal lines up with research comments in IAM-896 and makes sense at a first glance.  The redundant deletion is "I wonder why it was there" but it feels like .. if that breaks anything it's going to be a marvelous discovery, but it's probably garbage.  So, let's go.

PR #730, opened 2024-01-16
fixed url and added vanity url

vansterzzz:
# All PRs are assigned reviewers by default; selecting your own is optional.


gcoxmoz (approved):
IAM-1274

Reasonable followup to a recent deploy.

PR #735, opened 2024-01-23
added prefectv2 app

vansterzzz:
onboarding new app in IAM-1278 Prefect


gcoxmoz (approved):
Looks right to me, clearly flags the to-be-removed-later app for future cleanup (as suggested by the requestor).

PR #740, opened 2024-02-02
Upload only apps.yml and set cache control on objct

dividehex:
This PR changes the uploading of apps.yml.  We no longer sync the entire directory and instead only upload the `apps.yml` file.  We also apply the `Cache-Control: no-cache` header to the object.

### DIFF
--- a/.github/workflows/google-deploy.yaml
+++ b/.github/workflows/google-deploy.yaml
@@ -47,8 +47,10 @@ jobs:
         id: 'upload-to-bucket'
         uses: 'google-github-actions/upload-cloud-storage@v1'
         with:
-          path: './'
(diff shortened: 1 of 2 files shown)


dividehex:
Note: we currently upload apps.yml to both an AWS and GCP bucket.  This has been previously done in preparation for move from AWS to GCP.  In the future, when the cdn is shifted over to GCP we will no longer need to write to the AWS bucket.

gcoxmoz (approved):
Approving based on intent and a skim-read of syntax.  If there's bad syntax, you'll figure it out quickly, but the intent of "pare it down to just the file needed" makes sense and is worth r+ on "this looks right."

PR #745, opened 2024-02-13
updated url for prefect

vansterzzz:
IAM-1278 updating login url after meeting with prefect dev team


ctbfourone (approved):
+1, I get redirected to Auth0 to login so seems good to me

PR #749, opened 2024-02-28
updated authorized groups to team_moco

vansterzzz:


gcoxmoz (approved):
Lines up with what was asked for in IAM-1300.

PR #750, opened 2024-02-28
update logo images for cdn

ctbfourone:

So, I'm not clear why, but the prior Recorded Future logo images didn't get uploaded to the CDN. So this is kind of a stab in the dark hoping that they will this time, but if not I'll investigate further.


gcoxmoz (approved):
I don't know if it'll do anything for you, but I like it for clearer names, if nothing else.

PR #754, opened 2024-03-06
updated authorized groups for splunk

vansterzzz:
IAM-1301, updated authorized groups for splunk


gcoxmoz (approved):
Does what the bug says, looks right to me.

PR #755, opened 2024-03-07
updated braze app with new group and name

vansterzzz:
IAM-1279 removing references to pocket for Braze app and added new authorized groups.


ctbfourone (approved):
+1

PR #763, opened 2024-04-03
Add MZVC and MZAI to Navan and Jira Service Management [IAM-1318]

dividehex:
See IAM-1318


gcoxmoz (approved):
Lines up with what was asked for in the bug, simple group add.

PR #764, opened 2024-04-08
added miniOrange app

vansterzzz:
IAM-1322, added miniOrange app.


gcoxmoz (approved):
Seems right, as the issue mentions testing this out in foxtail (a dev wpe instance).

PR #765, opened 2024-04-19
added/updated miniorage apps to include stage and prod

vansterzzz:
IAM-1229 and IAM-1330: added stage, prod instances, and updated vanity URL.


gcoxmoz (approved):
It's IAM-1329 and IAM-1330 (typo in initial comment) but otherwise, replicates an already-working dev instance into stage and prod.

PR #766, opened 2024-04-24
updated salesforce/salescloud app name

vansterzzz:
IAM-1332: updated Salescloud app name to MoFo Salesforce


ctbfourone (approved):
+1

gcoxmoz (approved):
mofo person asking for a name change for a mofo-only service.  Seems legit.
Since the complaint is about tile prettiness, r+

Note that the vanity url is based on the going-away name.  I see no compelling directive to either change it or leave it, but just noting the desync here in case it's of concern.

PR #767, opened 2024-04-30
added common room app to dashboard

vansterzzz:
IAM-1339 adding common room app to dashboard.


ctbfourone (approved):
+1, looks like it matches IAM-1339 as requested

PR #772, opened 2024-05-15
updated chatgpt URL

vansterzzz:
IAM-1349 updating tile URL per vendor. Testing has been done with Mark/:freshness.


ctbfourone (approved):
+1, I don't have access but everything seems good.

PR #773, opened 2024-05-16
Add /gcal and /jsm vanity URLs

globau:
Adds a couple of vanity URLs that I guess incorrectly every time.


ctbfourone (approved):
+1, this looks good and super handy

PR #777, opened 2024-05-24
Adds temp travel user to all Jira entries with matching client_id

ctbfourone:

To hopefully address the issue with a travel account exception.


dividehex (approved):
👍

PR #779, opened 2024-05-29
added 4 Bedrock CMS apps

vansterzzz:
IAM-1350, added Bedrock CMS apps - prod, stage, dev, local dev


ctbfourone (approved):
+1, looks good and matches all the things

PR #784, opened 2024-06-14
Iam1364 1

vansterzzz:
IAM-1364, using a new file for the tile on dashboard. the previous one did not render/resize well.


gcoxmoz (approved):
Squareagons seem more in line with our tiles, untested but "sure, makes sense."

PR #785, opened 2024-06-17
IAM-1356: set up cloudservices aws sso app

dividehex:


gcoxmoz (approved):
Looks right vs what was asked for in the issue.

PR #787, opened 2024-06-18
updated tile image and name

vansterzzz:
IAM-1364, updating tile image and name per request for brand launch.


gcoxmoz (approved):
```
<gcox> Did you mean to change the name to ‘brand.mozilla’ (no .com)
<van> I noticed that's what she wanted in her dm
```
Future quibble on the moz logo being called frontify but approving to unblock.

PR #793, opened 2024-07-01
updated access group for grafana influx app

vansterzzz:
IAM-1378: added 2 access groups to grafana influx app earthangel.


ctbfourone:
My only concern is that the group name should be more specific since more than one team could use Grafana and want to hook it to SSO. Not a show stopper, just something to consider.

ctbfourone (approved):
+1!

PR #795, opened 2024-07-11
Slack payload message tweak

ctbfourone:
I noticed that ~`env.GITHUB_ACTOR`~ `env.RELEASE_NAME` is always `???` so I tossed that. Reordered the text fields a little, and made the values bold. And moved the workflow name to the empty text field spot.

### DIFF
--- a/.github/workflows/payload-slack-content.json
+++ b/.github/workflows/payload-slack-content.json
@@ -1,5 +1,5 @@
 {
-  "text": ":link-run:",
+  "text": ":triforce:",
   "attachments": [
     {
       "color": "{{ env.STATUS_COLOR }}",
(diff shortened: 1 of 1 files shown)


gcoxmoz (commented):
I don't quite follow this , but I spotted a nitpick.

> I noticed that env.GITHUB_ACTOR is always ??? so I tossed that.

You didn't toss this line, though.  You tossed `env.RELEASE_NAME`.   Doublecheck?

ctbfourone:
> I don't quite follow this , but I spotted a nitpick.
> 
> > I noticed that env.GITHUB_ACTOR is always ??? so I tossed that.
> 
> You didn't toss this line, though. You tossed `env.RELEASE_NAME`. Doublecheck?

Doh, sorry, good catch. The commit was right, but my note was wrong (edited it above).

gcoxmoz (approved):
Notes match intentions, I can see WHAT you're trying to fix in the messages from the slack channel.  Change makes sense, and if you're wrong the impact is localized to just the team, so SURE.

PR #796, opened 2024-07-11
Update LifeLabs URL

ctbfourone:
They require SP-initiated SSO, thus updating the url.


gcoxmoz (approved):
In service of IAM-1374.
URL change matches the ticket and was indicated as working.

PR #798, opened 2024-07-11
IAM-1384 Add Developer Services group to Cloudservices AWS SSO

jbuck:


gcoxmoz (approved):
Group name matches IAM-1384's discussion, is scoped to the RP, this makes sense.

PR #801, opened 2024-07-16
IAM-1389 Add FxA developers to Cloudservices AWS

jbuck:


gcoxmoz (approved):
Pattern-wise this follows previous art ala IAM-1384 so this seems fine, assuming the other 'make the group' stuff happens with it.

PR #802, opened 2024-07-18
Azure

dividehex:
Adds a client entry for Azure.   This is the first time adding a svg graphic and I'm not sure it will render but I want to find out.  If it doesn't, I'll convert it to png.


gcoxmoz (approved):
apps.yml change looks fine, experimental nature of trying svg is known and accepted as part of submitting this, lgtm.

PR #811, opened 2024-08-02
Add richer tests for YAML parsing, IAM-1399

gcoxmoz:
A deeper inspection of `apps.yaml` and trying to look for errors.  The regex's are tight (possibly too tight) but if they are then we can expand them in the future.


dividehex (approved):
Awesome! Thank you

PR #812, opened 2024-08-02
Remove old slack app, add vanity to new app, and display

dividehex:
This PR is the first step to the Slack migration.  It removes the old slack app, adds the vanity url to the new app and set the tile to display.   Before this PR is deployed to production, the old client will need to be deleted from Auth0.

old client_id: `WXVdgVoCca11OtpGlK8Ir3pR9CBAlSA5`


gcoxmoz (approved):
PR has captured the `authorized_groups` from the deleted entry, and calls out that is is a stepwise change and not a final change.  Makes sense.

PR #816, opened 2024-08-15
created new udemy app for mofo, removed mofo group access from moco udemy

vansterzzz:
IAM-1400: created Udemy app for MoFo and removed MoFo group access from Udemy - Business.


gcoxmoz (approved):
Van indicates that the loss of access to the old one by mofo is known and accepted.

PR #820, opened 2024-08-19
Restore nagios entry with new client id

dividehex:
See: IAM-1405


gcoxmoz (approved):
Matches the info from the ticket.  Sorry for the noise.

PR #821, opened 2024-08-19
Replace azure logo with new png image

dividehex:
The existing Azure image was a svg which doesn't render.  This replaces that file with a png file.


gcoxmoz (approved):
Fixes #802 's svg attempt.  "We tried!"

PR #828, opened 2024-09-05
updated URL for fundraiseup

vansterzzz:
IAM-1401, updated URL for FundRaise Up. Per their docs, they do not recommend IdP initiated logins.


gcoxmoz (approved):
I don't see where the new URL came from, but if the old one is busted then "you probably saw it" and "approving as breakfix".

ctbfourone (approved):
+1

PR #830, opened 2024-09-09
updated authorizzed groups for testrail app

vansterzzz:
IAM-1420: updated authorized groups for testrail to include team_mzla and testrail-access


gcoxmoz (approved):
Reasonable based on where that bug is going.

PR #832, opened 2024-09-10
added new webcompat fuzzing app

vansterzzz:
IAM-1421: added webcompat fuzzing app


gcoxmoz (approved):
New addition, reasonably tracks the bug, sure

PR #852, opened 2024-11-13
Add ADP (Non-SSO)

bheesham:
Jira: [IAM-1483](https://mozilla-hub.atlassian.net/browse/IAM-1483)


apps.yml
```diff
@@ -4117,3 +4117,16 @@ apps:
     url: https://auth.mozilla.auth0.com/samlp/cPH0znP4n74JvPf9Efc1w6O8KQWwT634
     vanity_url:
     - /tines
+- application:
+    authorized_groups:
```
gcoxmoz, line 4121: Maybe also `team_mzai` and `team_mzvc`?

bheesham: Added!

gcoxmoz (approved):
I THINK this is a yes.  From the notes at the top of `apps.yml` the groups are gating tile visibility to ADP based on groups, which is good.  Then, because there's no client_id we aren't restricting access to it.. which sounds "bad" but "it's just a glorified hyperlink" since it's not actually an SSO integration, so I don't see a problem there.  Seems to match intentions.

PR #856, opened 2024-12-02
Remove groups which have been replaced in peoplemo

bheesham:
Jira: IAM-1474


gcoxmoz (approved):
This is the desired path.  Worst I can imagine is someone gets culled when they didn't get moved to the new groups by accepting an invite.. too bad so sad for them, shipit.

dividehex (approved):
🍀 Approved

PR #857, opened 2024-12-02
updated fable authorized groups

vansterzzz:
IAM-1517: updated access for Fable app; left access group intact should individuals outside moco get invited.


gcoxmoz (approved):
Poked :van's brain about noodling over group cleanup on the the pmo group, but no blocking on that.

PR #866, opened 2025-01-15
added wiz app to sso dashboard

vansterzzz:
IAM-1532: added Wiz app including tile to SSO dashboard


bheesham (approved):
👍 Looks good to me!

PR #878, opened 2025-02-27
Add service_jira as an authorized group for Jira products, IAM-1580

gcoxmoz:
This supports non-moco users having a path to access, notably driving this is a test account.

### DIFF
--- a/apps.yml
+++ b/apps.yml
@@ -442,6 +442,7 @@ apps:
     - team_mzvc
     - team_mozillaonline
     - team_office_support
+    - service_jira
     - travelling_accounts_jira
     authorized_users: []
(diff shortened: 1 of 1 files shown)


bheesham (approved):
Looks good!

Though, one question: should we be adding them to `Jira - Stage` (line 421) as well? Their ticket mentions they're trying to add this user in "to test".

gcoxmoz:
Nope, because IAM-1579 will zorch that entry.

bheesham:
> Nope, because IAM-1579 will zorch that entry.

Ah, I missed that linked ticket.

In that case: 🚢 👁️ 🍵

PR #879, opened 2025-03-03
deleted jira-stage app

vansterzzz:
IAM-1579 delete jira-stage app


gcoxmoz (approved):
Application matches bug, lgtm, thanks!

PR #881, opened 2025-03-04
updated element/matrix client ID

vansterzzz:
IAM-1462: updated client ID for Matrix/Element chat


vansterzzz:
also updated the app name for more clarity.

ctbfourone (approved):
+1, matches Auth0 app, lgtm!

PR #887, opened 2025-03-11
Iam1589

vansterzzz:
IAM-1589: removed Rapid7 tile from SSO dashboard and deleted .png from /images.


gcoxmoz (approved):
In that the ticket is 'remove the tile' and not 'remove the client', this makes sense.

PR #892, opened 2025-03-24
removed voice from auth0

vansterzzz:
IAM-1570: decommission Common Voice and associated apps as they are no longer using auth0 IdP. Per dfeller: "We're routing all auth requests to Moz Accounts, so there is no need to keep those entries."


gcoxmoz (approved):
Removal of already-not-displaying items, tracking bug, no tile images to delete, crossreference doc from :van covers a wide scope.  Looks right.

bheesham (approved):

PR #896, opened 2025-04-01
removed team_mzvc from airmo

vansterzzz:
IAM-1604: removed team_mzvc from Air Mozilla.


gcoxmoz (approved):
Looks right, simple removal

bheesham (approved):

PR #901, opened 2025-04-08
Successful logins would result in redirection issue

bheesham:
Jira: [IAM-1608](https://mozilla-hub.atlassian.net/browse/IAM-1608)


gcoxmoz (approved):
What I see in the bug doesn't really explain this, but it's pretty small potatoes as changes go, so "sure!"

PR #903, opened 2025-04-17
added vevox

vansterzzz:
IAM-1620: added new Vevox app


gcoxmoz (approved):
New addition, seems reasonable

PR #904, opened 2025-04-21
Prepare to deny auth if application isn't listed

bheesham:

This isn't an ideal solution, as discussed previously. But, this aligns with the regular Auth0 administration workflow -- the IT/Infra team can continue to add applications without needing us to intervene.

We lose out on some security by obscurity, but... that might be fine? (Famous last words.)

---

See also:

* https://github.com/mozilla-iam/auth0-deploy/pull/506

### DIFF
--- a/.github/workflows/lint-apps.yml
+++ b/.github/workflows/lint-apps.yml
@@ -1,0 +1,45 @@
+name: Lint apps.yml
+
+on:
+  push:
+    branches:
+      - master
+  schedule:
+    - cron: "0 13 * * *"   # 9AM EST, 6AM PST, 1PM UTC
+
+jobs:
+  lint:
+    name: Lint against Auth0
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+      - name: Install dependencies
+        run: |
+          pip install -U pip
+          pip install -r requirements.txt
+      - name: Generate apps.new.yml
+        env:
+          AUTH0_DOMAIN: ${{ secrets.AUTH0_DOMAIN }}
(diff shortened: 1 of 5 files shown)


gcoxmoz (approved):
Right off the bat I can see things that should have been deleted, and I think this should spur some cleanup, but I don't see the point in holding up progress here; IMO 'if it's bad in auth0, but you faithfully replicate the state, so be it' and 'clean up later' isn't that much more complex if this lands.  So I'm assuming that your adds here is accurate 'as of now'.

bheesham:
@dividehex suggested (on Slack) we make a CI job to add a lint for this, so I'll do a bit of work on that.

dividehex (approved):
🍀 As discussed via Slack, clients in auth0 will be managed by Terraform in the near future.  This means we will be able to programmatically generate the apps.yml file.  In addition to that, we should put in some checks during the CI/CD process that query the auth0 API and bi-bidirectionally compare the tenant client list and the apps.yml list. **Approved**

dividehex:
So one issue that may not have been considered, is the apps.yml not only hosts app definitions for prod but also for dev.  Although, I don't think we want to lint against the dev tenant in parallel with prod, maybe we could annotate the dev application definitions so the linter can ignore them.  This would prevent them from being removed for the crime of not existing on the prod tenant. 😐

bheesham:
Ah, so that's the reason for [adding the dev SSO dashboard](https://github.com/mozilla-iam/sso-dashboard-configuration/pull/904/files#diff-996d6742fe48124b31444f5db29bfa7863c3155a4ca5a88173f1b17e9de558c0R10)!

I'll update the reasoning.

PR #905, opened 2025-04-23
Run pre-commit during gha ci/cd

dividehex:
This runs pre-commit in gha so if a pre-commit lint issue isn't caught before the devs commit on their localhost, it will catch it here.


bheesham (approved):

bheesham:
Solid!

PR #906, opened 2025-04-23
ci url output had trailing spaces, which fails yaml linting

bheesham:


gcoxmoz (approved):
You say it fails, I believe you.  Plus 'ew trailing spaces' so, bonus.

PR #907, opened 2025-04-23
Update `make test` to deal with schema change

bheesham:

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -16,4 +16,5 @@ jobs:
         python-version: 3.12
     - name: Run tests via Makefile
       run: |
+        pip install -r requirements.txt
         make test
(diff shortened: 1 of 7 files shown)


bheesham:
@gcoxmoz brought up a good point re: the schemas here -- `logo` and `url` are both set for regular apps. I've added them back here, though I believe they're optional.

We should probably have a larger discussion about schema validation, at some point, since there are now multiple consumers of this file.

gcoxmoz (approved):
Approving to unstick/breakfix.  This should help in codifying a schema near to what we already have, and someone can handsmack me if I'm overstepping here.

bheesham:
Forward the handsmacking to me, if anything -- I broke it. 🙃 

🎩 thanks for the approval.

PR #913, opened 2025-04-30
added pingdom app

vansterzzz:
IAM-1626: added pingdom app


gcoxmoz (approved):
Makes sense, lines up with the bug.

ctbfourone (approved):
+1

PR #917, opened 2025-05-12
Remove r7 and soar

densfox:
SEC-3823


bheesham (approved):
Apps deleted from Auth0. The PR looks good to me.

PR #919, opened 2025-05-19
adding new Navan integration

vansterzzz:
IAM-1646: new required Navan integration.


ctbfourone (approved):
I imagine you'll want drop "OIDC" from the name before merging so it doesn't confuse folks, but otherwise +1

PR #920, opened 2025-05-20
Remove team_pocket from Navan access

gcoxmoz:
bug 1707014 added `team_pocket` to egencia/tripactions/navan.  At the time (2021), `team_pocket` was all pocket people, and they were distinct from MoCo/MoFo.  Since then, Pocket is all integrated with MoCo, and as such this team group is unnecessary for Navan.


vansterzzz (approved):
LGTM +1

PR #921, opened 2025-05-20
decommissioned old navan app and updated URL and vanity for new navan…

vansterzzz:
IAM-1646 removed old Navan, changed URL and used tile for new Navan app.


gcoxmoz (approved):
That's a cutover, alright.  Looks right to me, assuming backend cleanup is done and URL is right.

PR #922, opened 2025-05-23
added vevox-participants app

vansterzzz:
IAM-1658: added Vevox-Participants. This is still a PoC. Will rename the other Vevox Admin app if we continue to go with this product.


bheesham (approved):
Looks good to me!

PR #926, opened 2025-06-23
removed Vevox admin and participant PoC apps

vansterzzz:
IAM-1691: removed Vevox admin and Vevox participant apps because PoC is over.


ctbfourone (approved):
+1

PR #954, opened 2025-09-10
updated access groups

vansterzzz:
IAM-1776: updating access groups for Lattice Prod and Lattice Sandbox


vansterzzz:
FYI tiles are still not visible until we roll out.

PR #962, opened 2025-10-01
set tile to visible for claude app

vansterzzz:
App is rolling out so we're going to make the tile visible to approved users.


gcoxmoz (approved):
flip dat switch

ctbmozilla-admin (approved):

bheesham (approved):

PR #1018, opened 2026-04-07
Add meao-admins to the cloudservices section

bheesham:
`mozilliansorg_meao-admins` was only mapped to IT. With @bwells-moz's PR we'll need to ensure it's mapped to CloudServices' as well.

See also: https://github.com/mozilla/global-platform-admin/pull/5872

Jira: SVCSE-4371


gcoxmoz (approved):
Makes sense, has a bug trail, goooo

PR #1023, opened 2026-04-15
removed mofo from original navan, created new navan for mofo

vansterzzz:
IAM-1946: removed MoFo from original Navan instance and created a new one for them.


ctbfourone (approved):
+1, LGTM

PR #1025, opened 2026-04-20
updated base44 URL

vansterzzz:
IAM-1937: updated Base44's URL to an RP initiated login.


gcoxmoz (approved):
URL change, sure why not

PR #1026, opened 2026-04-22
removed MoCo from Udemy

vansterzzz:
IAM-1956: removed MoCo from Udemy.


gcoxmoz (approved):
Removal of the whole moco-specific entry since we've closed out the contract, makes sense to me.

PR #1027, opened 2026-04-23
updated accelevants to an sso dashboard tile only

vansterzzz:
IAM-1925: remove accelevants from auth0, updated URL, and deleted the client ID in apps.yml. moved the application under ADP for reference because accelevants does not support auth0 but the requestor still wants a tile on the dashboard.

### DIFF
--- a/apps.yml
+++ b/apps.yml
@@ -3813,6 +3813,17 @@ apps:
     - /adp
 - application:
     authorized_groups:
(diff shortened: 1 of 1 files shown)


gcoxmoz (approved):
Drops the SSO aspect and removes the client_id debris, looks fine.

PR #1030, opened 2026-05-07
IAM-1949: Add Coursera

ctbfourone:

Adds the entry, hides it and has no vanity URL for now, but I'm trying to debug setting up the Coursera side and think since there's no apps.yml entry in prod I'm unable to even hit the sign-in url.


bheesham (approved):
🚀

PR #1034, opened 2026-05-12
fix(claude): IdP-initiated login not supported

bheesham:
The flow has to be SP-initiated for now. This may be a little confusing for users, since on the SP login page, they give an additional login option, Google.

Jira: IAM-1970


ctbmozilla-admin (dismissed):

ctbfourone (approved):
:shipit:

PR #1050, opened 2026-05-28
Cleanup exited users ; add Workato user [ITORG-200]

gcoxmoz:
Commit 1, a freebie-ridealong:  Removed two exited humans from `apps.yml`
Commit 2: Add `workato-automation@mozilla.com` - a nonhuman "LDAP shared_account" to Slack/Jira/Confluence.
This follows prior art (seen in the Confluence entry) of "using `authorized_users` instead of `authorized_groups` isn't great, but it should be fine for nonhumans / service accounts"

### DIFF
--- a/apps.yml
+++ b/apps.yml
@@ -2156,9 +2156,7 @@ apps:
     - team_moco
     - team_mofo
     - team_mzla
-    authorized_users:
-    - dlipski@mozilla.com
-    - jeide@mozilla.com
+    authorized_users: []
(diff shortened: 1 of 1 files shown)


bheesham (approved):
OH. MY. BLOB. I completely forgot about the `authorized_users` feature. 😮‍💨 

This'll work for these apps. This will likely also work for the other Workato apps too, which Mark asked about yesterday.

PR #1058, opened 2026-06-12
added 2 backstage.io apps

vansterzzz:
IAM-1987: added 2 backstage.io apps - prod and stage


gcoxmoz (approved):
New addition, hidden, sure have fun

bheesham (approved):

PR #1060, opened 2026-06-16
runlayer: enable for moco and mofo

bheesham:
Jira: IAM-1986

### DIFF
--- a/apps.yml
+++ b/apps.yml
@@ -6507,7 +6507,10 @@ apps:
     - /workato-workspace
 - application:
     authorized_groups:
+    - team_moco
+    - team_mofo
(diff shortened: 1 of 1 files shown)


gcoxmoz (approved):
Hacky seeing as this is an 'all the people' and a 'dedicated list' smooshed up together, but it's declared as 'needs a followup' so okay.

bheesham:
I don't mind the overlapping lists here. We can use that as a signal for which groups are actually mapped in the application.

There was some Custom Work :tm: required to get this set up properly on the RP's side, which includes specifying `mozilliansorg_mozillm-admins` somewhere in their backend.

PR #1061, opened 2026-06-17
fixed geo-tracker to have a stage and a prod instance

vansterzzz:
IAM-2000: fixed geo tracker prod and stage


gcoxmoz (approved):
New addition, makes sense.

bheesham (approved):

PR #1063, opened 2026-06-18
added new workday instance - implementation 1

vansterzzz:
IAM-2004: added a new workday instance - implementation 1


gcoxmoz (approved):
Follows the pattern of the other sandbox, LGTM

PR #1068, opened 2026-06-26
IAM-2007: Remove LifeLabs

ctbfourone:

Confirmed with Allison that this is good to go.


gcoxmoz (approved):
"... but I have read some obituaries with great pleasure."

PR #1077, opened 2026-07-08
Cleanup of team_elance and Coursera groups fix

gcoxmoz:
Mostly verbal, but some BIZSYS-1988 and IAM-1949.

Coursera missed getting all employees that should've been allowed in; half bad copyover from the Udemy decom, and half uncommunicated change in who was allowed.


ctbfourone (approved):
🦅

PR #1080, opened 2026-07-10
added coursera for MZLA

vansterzzz:
IAM-2019: added Coursera for MZLA


gcoxmoz (approved):
Parallel match to the other Coursera, lgtm

PR #1082, opened 2026-07-20
feat: Add test RPs

bheesham:
I spun up some test RPs using authlib/auth-playground deployed to GCPv2. These give us a safe space to test apps with high assurance requirements.

Jira: IAM-1989


gcoxmoz (approved):
Test instances, none displayed, none with a vanity url, linter makes sense.. what could go wrong?

kgowrabathuni (approved):

PR #1084, opened 2026-07-20
Added new workato workspace groups

kgowrabathuni:

IAM-2033: Adding new Workato workspace groups.

### DIFF
--- a/apps.yml
+++ b/apps.yml
@@ -6508,7 +6508,6 @@ apps:
     - mozilliansorg_workato_workspace_group-financesystems-viewonly
     - mozilliansorg_workato_workspace_group-financesystems-devs
     - mozilliansorg_workato_workspace_group-financesystems-admins
-
     authorized_users: []
     client_id: JmJAOmGbtZsojMpFQC5fcmqghWHbuKrf
     display: false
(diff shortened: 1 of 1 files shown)


apps.yml
```diff
@@ -6507,6 +6495,19 @@ apps:
     - mozilliansorg_workato_workspace_role-test
     - mozilliansorg_workato_workspace_role-prod
     - mozilliansorg_workato_workspace_group-mozilla-product-management
+    - mozilliansorg_workato_workspace_group-wptatomations-admins
```
gcoxmoz: `atomations` feels like a typo

gcoxmoz (changes requested):
Very likely typo spotted.
Also, I wouldn't reject based on this, but since I'm requesting changes, wouldn't be upset if the groups were put in something closer to alphabetical order.  Makes eyeball-reads easier later, but if there's a good reason not to, then disregard.

apps.yml
kgowrabathuni: ahh.. thanks Greg! Corrected the typo. We are following the latest groups added to the end.

gcoxmoz (approved):
Not sure why these couldn't still get sorted and still be at the end but okay.

PR #1087, opened 2026-07-21
LDAP group cleanup

gcoxmoz:

This is a wad of LDAP groups that have disappeared over time and not been cleaned up.

Assuming approved, there is no need to force a release/deploy, it can ride on 'the next deploy'.


ctbfourone (approved):
🧹

PR #1090, opened 2026-07-22
Remove ConsoleFlow

gcoxmoz:

Creation was [bug 1576983](https://bugzilla.mozilla.org/show_bug.cgi?id=1576983)
<johnb> :dividehex wasn't able to filter the groups down to the one needed so Lantronix would never give someone access consistently.  Only the few times the right group was the first in the list.

Please zap auth0, no rush on a deploy afterwards since it's just a cleanup.

### DIFF
--- a/apps.yml
+++ b/apps.yml
@@ -2173,17 +2173,6 @@ apps:
     op: auth0
     url: https://recit.getpocket.dev/login
 - application:
-    authorized_groups:
-    - team_netops
-    - team_infra
-    authorized_users: []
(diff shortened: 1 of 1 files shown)


kgowrabathuni (approved):

bheesham (approved):

bheesham:
Piped ConsoleFlow (`OSKMKPkvUoDg0K8agd4iHSwUam0lpr6p`) into `/dev/null` on our Auth0 machine. Thanks!

PR #1092, opened 2026-07-31
docs: Clarify what happens when an app specifies AAL but we require MFA

bheesham:
Spawned from https://github.com/mozilla-iam/auth0-deploy/pull/545#discussion_r3686213087.

Jira: [IAM-1989](https://mozilla-hub.atlassian.net/browse/IAM-1989)

### DIFF
--- a/apps.yml
+++ b/apps.yml
@@ -29,7 +29,15 @@
 #
 # 1. If the client_id is missing, MFA is not required. (This is uncommon.)
 #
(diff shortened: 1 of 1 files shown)


gcoxmoz (approved):
Thanks.  This is plenty of breadcrumbs to uncover that 'something is different' and if it ever needs wordsmithing to be more accurate, that's later's problem.

PR #1094, opened 2026-08-11
Add Anthropic for mzai

bheesham:
Jira: IAM-2067

- [X] **New integrations:** Legal _and_ Security reviews confirmed. `authorized_groups` and Auth0 `client_id` are defined. If `display: true`, the logo's image is attached. Auth0 app's Connections enables LDAP only.

Mario said:

> Our processes are way different from MoCo ones
> From our side, all is approved, and ready


gcoxmoz (approved):
Lil OCD twitch at 'claudemofo' but not enough to throw a flag on it.

PR #1100, opened 2026-08-21
Make governance a bit easier

bheesham:
The same app having different risk levels is sort of... annoying to keep track of.

Jira: IAM-1989


.git-blame-ignore-revs
```diff
@@ -0,0 +1 @@
+75f405d977865390cf06232d510ed70276a7f407
```
bheesham: blech, i'll remove this

gcoxmoz (approved):
Love a Linter.

PR #1106, opened 2026-08-27
updated icons for workday

vansterzzz:
IAM-2089: added new icon for workday-sandbox


gcoxmoz (approved):
Can't be mad about a differentiator, good fix.

PR #1109, opened 2026-09-01
Move Stripe groups to PMO

bheesham:
Jira: [IAM-2043](https://mozilla-hub.atlassian.net/browse/IAM-2043)


gcoxmoz (approved):
So, approved, but with a future caution.
If you pull old-groups and add new-groups, someone who hasn't auth0-refreshed may lack access in the new group for a while (30d?).   This might should be "add the new groups, wait 30d, pull the old groups".  But "hey log out and back in" can work around it.
Mostly a note for bigger moves later.

PR #1110, opened 2026-09-02
updated access group for workday preview

vansterzzz:
IAM-2100: updated access group for workday preview


gcoxmoz (approved):
Makes sense to narrow the scope.

PR #1111, opened 2026-09-03
added vanity URLs for workday

vansterzzz:
IAM-2103: added Vanity URLs for WorkDay app.


gcoxmoz (approved):
Additions to already-existing RPs, names match, lgtm

PR #1113, opened 2026-09-08
Added MOFO group for greenhouse moco access (temp)

kgowrabathuni:

IAM-2085: Greenhouse support team requested for a temporary access to Greenhouse Moco for MoFo users until Greenhouse MoFo SSO setup is complete.


gcoxmoz (approved):
Next stanza isn't fully plumbed yet so the temp here is to bridge mofo to the new instance.

PR #1114, opened 2026-09-08
updated access group for Firefox Prog Dashboard

vansterzzz:
IAM-2105: updated access group for Den Firefox Program Dashboard


gcoxmoz (approved):
semi-conflicting request for 'all staff' and 'moco', but they did say moco so okay.

PR #1121, opened 2026-09-18
Limiting access to MoFo recruiting only

kgowrabathuni:

IAM-2085: MoFo recruiting team requested liming Greenhouse (MoFo) access to only recruiting team until they complete the implementation.


gcoxmoz (approved):
Not loving a human-user list, but I'm assuming this will be reverted, so I can see why you wouldn't want to make a temp group.  Okay.

